PR #49, opened 2015-08-08
removed redundant </ul><ul> between <li>'s

otjura:
I really wonder if anyone seriously uses anything else except SLIME these days.


eudoxia0:
Thanks!

PR #50, opened 2015-08-08
expanded readme.md

otjura:
Not sure if I got the project goals right. But every project needs goals.


eudoxia0:
Approximate goals are better than none after all.

pnathan:
:+1:

Issue #37, opened 2015-08-06
Moving to Markdown


eudoxia0:
After #35 is done, we start turning the HTML into Markdown.

This is pretty extensive so any help is appreciated.

chuchana:
I am currently converting 'hashes'.

chuchana:
working of 'functions'

eudoxia0:
Protip: http://domchristie.github.io/to-markdown/

chuchana:
> Protip: http://domchristie.github.io/to-markdown/

Thank you. I knew should have been looking for something like that…

chuchana:
working on "systems"

chuchana:
working on misc

chuchana:
working on "emacs"

chuchana:
working on "macros"

chuchana:
working on "process" chapter

chuchana:
working on "windows"

chuchana:
working on "win32"

chuchana:
done!

pnathan:
@chuchana - You *totally* rocked this. :yellow_heart:

eudoxia0:
Good job team

PR #53, opened 2015-08-09
[Refs #37] Move the "emacs" chapter to Markdown

chuchana:
Maybe have a look at the re-indented .lisp files: I auto-indented them (with `(setq indent-tabs-mode nil)` so I can copy to the markdown files), but I am not sure whether the indentation for `loop` is correct.

On my local installation of Jekyll, the links to the lisp files (in _includes/code/) do not work although the rendered relative path is the same as the one on the github page (i.e. …/cl-cookbook/s1.lisp). The CSS doesn't work locally either, so hope it works on github.


eudoxia0:
The idea of putting them in `_includes/` is so we can do:

```
{% include "code/s12.lisp" %}
```

And include them textually without copying them. This prevents the files from going out of sync with the text.

chuchana:
It works (but without the quotes). I'll finish and create a pull request.

PR #55, opened 2015-08-09
[Refs #37] Move the "macros" chapter to Markdown

chuchana:
I have reinserted html code where it could not be represented in markdown and removed the strong emphasis of the backquote. I had enough work on that section without it and it wasn't recognisable in the html version anyway.


eudoxia0:
Excellent work!

PR #56, opened 2015-08-10
moved 'os', 'process', 'windows' and 'w32' to md

chuchana:
The process/threads chapter is Lispworks-specific. --> [Bordeaux Threads](https://common-lisp.net/project/bordeaux-threads/)?


pnathan:
@chuchana - Yes. The existing chapter should probably (I only quick-skimmed it) be migrated to LispWorks, and a new chapter started for using Bordeaux Threads.

chuchana:
@pnathan I tried starting to port it without any previous knowledge of Bordeaux-Threads, but I tripped very near the beginning: 
> Before you go any further, you might want to locate the "Process Browser". Aptly enough, you'll find it on the LispWorks podium under the icon of somebody running away. Open one **before** you get into trouble. To kill a process which won't respond to polite coercion, select it in the Process Browser and then click the "skull and crossbones" icon.

I found how to kill a thread that I have assigned to a variable:
` (setf *foo* (bt:make-thread (lambda () (loop)) :name "Foo2"))` --> `(bt:destroy-thread *foo*)`
but not how to kill just any thread that I can see with `(bt:all-threads)`:
`(bt:make-thread (lambda () (loop)) :name "Foo")` --> ?

That has got to be possible too, right?

eudoxia0:
We can investigate that later, we should move to markdown before making too many edits to the content. One step at a time.

chuchana:
I just started locally in a new file, the content I pushed is all original. Is there anything else I can do for the move to markdown?

pnathan:
quick skim LGTM.

PR #63, opened 2017-05-18
fix getenv on os page

vindarel:
- change outdated code snippet with example given on issue 3,
- tell about uiop:get-env as suggested in the issue,
- a word about osicat


eudoxia0:
Thanks!

PR #70, opened 2017-06-21
Create license info

pnathan:
Port over the License file, which apparently got munched at some point in the migration.

/cc @vindarel @eudoxia0 @LispCookbook/owners


pnathan:
gah, failure to read text on my part. :-X

PR #73, opened 2017-06-23
hash table: a word about CL21

vindarel:


pnathan:
I'd suggest something like "If we are using the CL21 extension library..."

vindarel:
hello, and now ?

PR #76, opened 2017-06-23
Portacle

vindarel:


pnathan:
Hm. Waving the wand of merging over this, but my instinct is that a smoother wordsmithing could be done so that the language flows well.

PR #79, opened 2017-06-26
Fix Other Resources

vindarel:
I removed dead links and added a few modern ones at the top.

The last link "dynamiclearningcenter.com" redirected to FranzInc and I kept it.


pnathan:
Looks like Norvig's style guide is here: https://www.cs.umd.edu/~nau/cmsc421/norvig-lisp-style.pdf

which is a valuable PDF

PR #80, opened 2017-06-26
New page Editors support, with an Atom plugin

vindarel:
which links to the two old pages and
to Articulate CL that references plugins for Vim, Atom and
others. The Atom plugin is newish and great for beginners. Shall we
copy content into this page so it is directly accessible ?

(I also accept some wordsmithing :) )

### DIFF
--- a/editors-support.md
+++ b/editors-support.md
@@ -1,0 +1,20 @@
+---
+title: Editors support
+---
+
+The editor of choice is still Emacs, but it is not the only one.
+
+## Editors plugins
+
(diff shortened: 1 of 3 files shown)


editors-support.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Editors support
```
pnathan: s/Editors/Editor/ support. Not sure certain which grammar rule this is (indefinite plurals?), but the noun goes in singular form.

PR #82, opened 2017-06-27
Files: get files attributes (size, mtime) with Osicat-posix

vindarel:
All I could find on the net or on Practical CL or on CL Recipes was about doing

```
(with-open-file (file "some file name") 
  (file-length file))
```

and nothing for other file attributes. I had to ask on SO, osicat-posix doesn't even have documentation O_o (while osicat does have one) (I opened an issue).

### DIFF
--- a/files.md
+++ b/files.md
@@ -292,3 +292,21 @@ CL-USER> (with-input-from-string (stream "I'm not amused")
 #\n
 5
 ~~~
+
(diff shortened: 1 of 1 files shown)


pnathan:
Jeeze.

It's aggravating how fiddly it is to interop with the outside world in CL.

PR #86, opened 2017-06-30
Strings libs

vindarel:


pnathan:
Looks great.

This is the page I use absolutely the most in the Cookbook.

I wonder how useful a giant "include all the above" library for string handling would be. I notice you have 3 libraries listed.... why not just one that Does All The String? But that's unrelated to the Cookbook.

vindarel:
Exactly, exactly, exactly. I too look at this page frequently. Basic string manipulation was the first thing I needed to do and so it was very frustrating. I googled a lot, gathered some stuff and created my mininal `str`. Only right after did I find `cl-strings`, I don't remember how, and referenced it in a couple resources. Indeed it would totally make sense there was only one (maybe except cl-change-case). It will depend on our lisp usage, motivation, and feedback. As usual I guess.

cl-strings and str are a bit different though. My str actually heavily uses other third-party libraries (cl-ppcre), when cl-strings re-implements the functionality, with potential issues discussed here: https://github.com/diogoalexandrefranco/cl-strings/issues/2 (remember ?) of which the portability issue seems to be fixed. TLDR;  `str` wants to be simple(r).

Issue #89, opened 2017-07-12
Old process/threads page, no mention of Bordeaux Threads


vindarel:
The process/threads page (https://lispcookbook.github.io/cl-cookbook/process.html) only refers to LispWorks:

> refers only to LispWorks. It has been tested on LispWorks for Windows Professional version 4.2.6 running on Windows NT (SP 6). All the examples below (apart from those few which require resaving the lisp image) should work without difficulty on the (free) Personal edition running under either Linux or Windows.

> Your mileage may vary. I will add “support” for other implementations and operating systems to this page later, if I have the time and access to the appropriate materials.

it doesn't mention the multiplatform Bordeaux Threads library.

vindarel:
I've had also good experience with lparallel: https://lparallel.org/

unfortunately https://github.com/orthecreedence/cl-async looks unfinished.

pnathan:
Yes, I think bordeaux-threads is Standard at the present time, and lparallel has good reports; I think it uses bordeaux-threads? Dunno.

Probably the correct fix is have a sequence of b-t snippets to create, join, block, destroy, etc, along with any lparallel snippets.

vindarel:
> I think it uses bordeaux-threads?

you're right (so says this lparallel tutorial https://z0ltan.wordpress.com/2016/09/09/basic-concurrency-and-parallelism-in-common-lisp-part-4a-parallelism-using-lparallel-fundamentals/#channels)

So what shall I do ?
- delete this content
- reference the libraries
- ask the author of https://z0ltan.wordpress.com/2016/09/02/basic-concurrency-and-parallelism-in-common-lisp-part-3-concurrency-using-bordeaux-and-sbcl-threads/ to take bits of his posts ? (I would make them shorter and add some stuff, like the `pmap` function I used in the web scraper page).

pnathan:
Probably ask the author. 

What I myself would do would write a small program with lparallel and use those snippets as the demo.

timmyjose:
Please feel free to use any materials as you see fit from my blog! I am quite flattered that you consider that material good enough for use in the cookbook!

timmyjose:
(I am the author of that blog).

vindarel:
That's great thank you ! So you're welcome to write on CFFI, a GUI toolkit, databases, maths, types and speed, lisp on mobile, or anything else :D

timmyjose:
Excellent! :-)

PR #90, opened 2017-07-12
process: add a note on Bordeaux Threads and a blog post

vindarel:
for #89 

just a side note and a link to a blog post, instead of re-writing everything.


pnathan:
:+1:

PR #95, opened 2017-07-17
index: TOC and contributors lists in 2 and 4 columns

vindarel:
A jquery and CSS trick to put the global TOC on the index on two columns and the contributors list on four, so that we have the whole toc at a glance and less whitespace, less scrolling. Looks better to me:

selection_163

selection_164


pnathan:
:+1:

PR #97, opened 2017-07-17
Cl21

vindarel:
Pushing for CL21 with an entry at the top.

CL21 is great !

### DIFF
--- a/cl21.md
+++ b/cl21.md
@@ -1,0 +1,358 @@
+---
+title: Common Lisp for the 21st century
+---
+
+[CL21](http://cl21.org/) is an experimental project redesigning Common Lisp. It features:
+
+* more functional programming facilities,
+* more object oriented, generic functions.
+* new syntax (for regular expressions, hash-tables, string-interpolation,…),
+* symbols organized into several packages,
+* written in pure Common Lisp.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


vindarel:
read it online: https://vindarel.github.io/cl-cookbook/cl21.html

pnathan:
is cl21 that good in pracice? It didn't appeal to me when I first head about it...

vindarel:
Open question since I didn't write large projects. But I really like the map for lists and vectors (instead of `mapcar 'vector`, generic stuff, shorter lambdas, … and everything is appealing to me :p

pnathan:
Let's put this under an "experimental" page heading then until someone
whose used it in a bigger project steps in to remark on it. I get the "just
appealing" factor, but I don't think we should recommend it (since that is
effect what the Cookbook will be doing) as a normative thing.

vindarel:
I'm really keen on advertising this, but I reckon we may not recommend it in a normative thing, for another reason that it is not guaranteed to be stable. I suggest this: we keep it at the top of the index for better visibility, but I changed its link title to "Experimental CL21" (instead of "modern Common Lisp").

vindarel:
Merged, but I can change something if needed :)

PR #112, opened 2017-09-17
justifying text with format directives

vindarel:
Documented some format directives in order to print and justify a list of strings, added ones to change case.

like this:

     1 Matrix                 5 
    10 Matrix Trilogy swe sub 3.3 

https://vindarel.github.io/cl-cookbook/strings.html#string-formatting

Comments and suggestions appreciated !

*edit*: you can also try out a fixed and scrollable sidebar.


pnathan:
Niiiice

PR #122, opened 2017-11-21
fix little error in demo

imnisen:
now the until macro 'will simply loops around until the condition becomes true' as the doc says


vindarel:
Good catch ! Ping to @timmyjose (hello!), the original author who has this on his blog too.

timmyjose:
Thanks, @imnisen @vindarel for spotting the error! I have updated my blog which contained those snippets as well! :-)

PR #126, opened 2017-12-02
getting started with Docker

vindarel:
https://github.com/daewok/lisp-devel-docker ships recent versions of SBCL, CCL, ECL, ABCL, and Quicklisp.

It was by far the easiest way for me to use a Lisp on my VPS (when SBCL can not start…).


pnathan:
*MB, no Mo.

Looks good.

vindarel:
Thanks ! Corrected and rebased.

Issue #41, opened 2015-08-06
Split the pattern matching and regular expression sections


eudoxia0:
Pattern matching should discuss optima and regular expressions should discuss cl-ppcre

PuercoPop:
Consider at lesat mentioning [trivia](https://github.com/guicho271828/trivia). It has the same API than optima and share tests. But it is easier to extend through defpattern.

PR #130, opened 2017-12-21
split regular expressions from pattern matching, as required in #41

vityok:
split pattern matching and regular expressions apart. please check if everything is fine, especially with git commit history, because git command line is a cryptic unusable counterintuitive mess and newbies do make mistakes using it.

this is to fix issue #41.


vindarel:
That's fine, thanks. You just seem to not have configured your email properly, because github doesn't recognize you as the commit author (shows a grey icon). You might want

    git config [--local] user.email <email>

vityok:
Hi @vindarel 

thanks for the tip. Hope it works now.

With best regards

vindarel:
It worked but it was not the cleanest. I rebased them into one with you as the author (`Vityok <bobbie@ua.fm>`) (but oops I didn't get it the first time, had to retry and push `-f`).

PR #132, opened 2017-12-21
Error and condition handling

vindarel:
This is what I came up with so far. I'm still new to this so please review carefully ! 

https://vindarel.github.io/cl-cookbook/error_handling.html

Known limitations:

- The introduction could be better
- I see two remaining "todos": explain the difference between `error` and `simple-error` (and others?),
- explain `make-condition` vs `define-condition`: needed ?

This is the basis I'd have loved to read, please tell me if I can add more important stuff.

fixes #131 - Request: error and condition handling (PR inside)

PR #135, opened 2017-12-26
Scripting page: parsing CLI args, building self-contained executables

vindarel:
read online: https://vindarel.github.io/cl-cookbook/scripting.html

with one reference to the Error handling PR: https://github.com/LispCookbook/cl-cookbook/pull/132 (which needs more care)

### DIFF
--- a/index.md
+++ b/index.md
@@ -32,6 +32,7 @@ title: Home
 * [Threads](process.html)
 * [Defining Systems](systems.html)
 * [Using the Win32 API](win32.html)
+* [Scripting. Building executables](scripting.html)
 * [Testing and Continuous Integration](testing.html)
 * [Web Scraping](web-scraping.html)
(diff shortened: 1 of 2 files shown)


vityok:
Great recipe, but it might benefit from using portable non-SBCL specific APIs.

For example, instead of `sb-thread` use `bordeaux-threads` (only difference is with `bt:all-threads` instead of `sb-thread:list-all-threads`).

Also, graceful exit from the Lisp system can be accomplished by `uiop:quit` (probably could be mentioned there).

vindarel:
Thanks, just what I needed. Any opinion on the other condition handling PR ?

pnathan:
This is good. I'd be curious as to actual emitted code size vs a self-contained JRE+uberjar, as that's effectively the same sort of thing.

PR #136, opened 2017-12-31
Structs

vindarel:


data-structures.md
```diff
@@ -1264,6 +1264,182 @@ we remove an element with `remf`.
 ;; => "foo!!!"
 ~~~
 
+## Structures
+
+Structures offer a way to store data in named slots. They support
+single inheritance.
+
+They can be considered super-seeded by classes of the Common Lisp
```
pnathan: superceded

data-structures.md
```diff
@@ -1264,6 +1264,182 @@ we remove an element with `remf`.
...
+Object System (CLOS), that don't have their limitations (see below).
```
pnathan: Also, DEFMETHOD doesn't dispatch on DEFSTRUCT (iirc - might need to test that assertion, but it's what I recollect).  DEFMETHOD is slower than DEFUN though, so... ymmv.

And, structs don't really interop with clos too well.

vindarel: Thanks. I don't feel like talking about defmethod here (it wasn't introduced yet), 

> And, structs don't really interop with clos too well.

I invite you to rephrase what can be added :)

pnathan: that's fine.

pnathan:
:+1:

PR #141, opened 2018-01-02
Detailed instructions on initial setup

can3p:
Github uses customized jekyll setup, that goes with bundled dependencies their own markdown implementation etc, so sometimes it's easier to use it instead of stock jekyll.

In addition to the the whole ruby/gems thing can be tricky due to dependencies mismatch and similar thing, so I've added a section to readme with detailed instructions, to make contributing easier.


pnathan:
please fix conflicts

pnathan:
(this is also a great idea and I will merge once the git issue is untangled)

can3p:
I resolved the conflict using github web interface and now it still shows pull request as broken on this page but as resolve on the page of my fork. What should I do to solve it there as well?

pnathan:
curious. <3 git

pnathan:
:sigh:

pnathan:
thank you for your contribution

PR #144, opened 2018-01-03
Improved grammar and information regarding structures.

jsjolen:
Hi!

I've swapped "loose" for "lose" (since that is correct grammar :-) ).

I removed the comment about MAKE-INSTANCE since it is indeed correct CL to create STRUCTURE-OBJECTs through it!

I also tried to clear up the phrasing of the lack of introspection available to structures.

### DIFF
--- a/data-structures.md
+++ b/data-structures.md
@@ -1269,8 +1269,7 @@ we remove an element with `remf`.
 Structures offer a way to store data in named slots. They support
 single inheritance.
 
-They can be considered superceded by classes of the Common Lisp
-Object System (CLOS), that don't have their limitations (see below).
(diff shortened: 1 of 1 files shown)


vindarel:
Absolutely ! Thanks.

jsjolen:
@vindarel , I decided to do some more research regarding make-instance.

Both SBCL and ECL supports using make-instance on structure-classes but this doesn't seem to be supported according to the standard (see links below). SBCL and ECL defines a method which specializes on CLASS, which is why it's supported in their respective impl.s.

I'm sorry for supplying a bad PR :-/.

http://www.lispworks.com/documentation/HyperSpec/Body/f_mk_ins.htm

http://www.lispworks.com/documentation/HyperSpec/Body/07_ag.htm

vindarel:
No don't worry it was all good !

PR #145, opened 2018-01-03
CSS: distinguish link to page source from page content

vindarel:
Some CSS around "page source", to distinguish it from page content and make it even more attractive.

selection_199

(welcome to designers)


can3p:
Looks much better!

PR #153, opened 2018-01-16
Files: read a file with uiop (no Alexandria), write to file

vindarel:
Just learned about uiop:read-file-[string,lines], so no need of Alexandria for that.

Added quick section "write to file".


pnathan:
nice

PR #160, opened 2018-02-20
Fix two links in functions.md

bendudson:
Two markdown syntax errors, and one broken link replaced by Wikipedia link (currying).


vindarel:
thank you !

Issue #149, opened 2018-01-09
Multidimensional arrays


bendudson:
I would like to make a start at writing a page on arrays in Common Lisp, aiming at something like this https://docs.julialang.org/en/stable/manual/arrays/ . I am interested in numerical/scientific computing, mainly solving PDEs from plasma physics, but would like to help this project if possible.  Topics would include:

- Types of arrays and vectors, including native and foreign arrays
- Operations on arrays, including element-wise and contraction
- Slicing and displacements
- Broadcasting 

Practical Common Lisp has a section on vectors and arrays, but does not go into much detail or cover much of this. I think most of this functionality is provided by libraries rather than the core language. 

I am very new to CL, so I plan to write a guide for my own benefit anyway. This is my impression, apologies in advance for errors or omissions:

There seem to be some quite mature libraries like Antik and GSLL, though the manual link here https://www.common-lisp.net/project/antik/index.html is broken, and documentation in general seems to be sparse. The other active(ish) library seems to be matlisp ( https://github.com/matlisp/matlisp ), though I am not sure how well these packages interact.

Several other projects exist, but most seem to be abandoned. Lisplab has a manual (https://common-lisp.net/project/lisplab/) but it's not on quicklisp. Other interesting but abandoned projects include http://quickdocs.org/array-operations/ which is on github (https://github.com/tpapp/array-operations). 

A relevant question on SO is: https://stackoverflow.com/questions/4116069/lisp-multidimensional-array-elementwise-operations

Is there a good guide to array operations in CL which I have missed? I understand that technical computing is not a focus for CL, but perhaps putting together a guide for new users would help consolidate the various libraries out there. Any suggestions or starting points welcome.

vindarel:
Hello, I myself can absolutely not tell you if there's a good guide about this out there :] but I can encourage you to write about it, yes, please write about this and about math in CL in general (issue about math: https://github.com/LispCookbook/cl-cookbook/issues/17).

Though, is [Maxima](http://maxima.sourceforge.net/documentation.html) not helpful ? algebra system, 

>  system for the manipulation of symbolic and numerical expressions, including differentiation, integration, Taylor series, Laplace transforms, ordinary differential equations, systems of linear equations, polynomials, sets, lists, **vectors** <--- \o/, matrices and tensors.

Also just stumbled across https://github.com/sjl/vex/ "yet another vector library", not ready but by an awesome CL hacker, to whom you could ask. 

[...]

PR #162, opened 2018-02-21
Multidimensional arrays page

bendudson:
This partly addresses issue #149, though is probably not ready yet. 

I have made a collection of useful recipes for manipulating standard arrays, rather than choosing one of the many CLOS-based extension libraries. I hope these are more generically useful.

Some of the libraries I've used, array-operations and LLA, have been abandoned by the author (https://tpapp.github.io/post/orphaned-lisp-libraries/), but they were the most convenient, or at least easiest to get working, of the many I looked at. I have adopted array-operations (https://github.com/bendudson/array-operations) and added to it, and will try to improve the documentation for LLA when I figure out how it works. 

Some of the array-operations functions may be found elsewhere, for example `each` is essentially the same as `array-map` from https://stackoverflow.com/questions/4116069/lisp-multidimensional-array-elementwise-operations . I thought it would be better to use a Quicklisp library where possible.

I've included some small utility functions and macros, which I found very useful, and couldn't find elsewhere. Is that ok? They are slightly simplified versions of things I've added to array-operations, but are not in the quicklisp version. I am not very experienced in CL, so these are likely not optimal or in the best style. Comments and improvements welcome!


vindarel:
Wooo, fantastic resource ! 

> I've included some small utility functions and macros, which I found very useful, and couldn't find elsewhere. Is that ok?

it's ok, and it will be even better when you find a place for them in a QL library :)

About array-operations, I'm sure you could become the new maintainer, did you ask ? He can give you push access or transfer ownership, and you'd have to notify QL by a new issue.

vindarel:
consequently, you could add a note in those snippets that they are in your version of array-operations (they are only a git clone in ql/local-projects after all), generally it's ok to me to advertise your active fork.

vindarel:
Also no mention of CLASP ? Just curious.

bendudson:
I haven't asked him yet, as I haven't been maintaining it long, but it's a good idea and I should ask.
CLASP would be good to add, but I haven't any experience (yet) of using it. It currently requires a compile from source so is probably not very beginner-friendly. I can definitely add a link to it.

PR #163, opened 2018-02-23
Improvements to multidimensional arrays page

bendudson:
Sorry, just too late... I think I've improved it in a few ways

- Mentions MAGICL and CLASP. MAGICL doesn't work directly with native arrays, so I have not included examples yet
- Added notes about array-operations repository
- Fixed (I think) my naive implementation of nested-loops, which used eval and would not work in general. The new version fixes these issues and has been (slightly) more thoroughly tested.
- Expanded some of the loop examples


vindarel:
sorry too, I'm impatient :)

PR #164, opened 2018-02-25
Link to modern public domain online CLOS MOP specification.

Hexstream:
[Almost entirely copy/pasted from [a similar pull request on awesome-cl](https://github.com/CodyReichert/awesome-cl/pull/124).]

Hello,

My modern public domain online version of the CLOS MOP specification was first released on 15 october 2017, and after further improvements, although my full vision has not yet been realized I feel this is in quite good enough shape to submit to the "Other CL Resources" section of the Common Lisp cookbook.

Full faithfulness to the contents of the book has been a top priority, which was ensured by carefully auditing the new version against the actual book, so this is very complete and safe to use. Mobile support is already implemented, and offline support and further improvements are underway.

I think it's safe to say that most people would prefer to use this new version over any other previously existing versions. A cursory look at my version should easily convince most people of this.

https://clos-mop.hexstreamsoft.com/

I hope this is to your satisfaction.

-------------------------------------------

(I inserted my entry right after the CLHS and CLUS entries because the CLOS MOP is probably the most important Common Lisp de-facto standard, and it is widely supported in all modern Common Lisp implementations, and users can easily make use of it through the `closer-mop` de-facto standard compatibility library (which was a Top 3 Common Lisp library for all of 2017 according to Quicklisp download stats). (I also wonder in passing if the CLUS entry should be a bit lower due to incompleteness, but I don't want to be a dick about it.))

(I added the "(modern public domain version)" part because I think quite a lot of people, when seeing "CLOS MOP specification", would think this is just either the old copyrighted no-modifications-allowed version or the Robert Strandh version.)

(Note that all or almost all of the CSS validation "errors" (as seen when clicking on "✔ CSS3" at the bottom of the page) stem from the validator not yet supporting CSS variables, a widely supported feature in all modern standards-compliant browsers. I consider this a bug in the validator.)

### DIFF
--- a/index.md
+++ b/index.md
@@ -46,6 +46,7 @@ title: Home
 * The [Awesome-cl](https://github.com/CodyReichert/awesome-cl) list
 * [The Common Lisp HyperSpec](http://www.lispworks.com/documentation/HyperSpec/Front/index.htm) by Kent M. Pitman
 * [The Common Lisp UltraSpec](http://phoe.tymoon.eu/clus/doku.php)
+* [CLOS MOP specification](https://clos-mop.hexstreamsoft.com/) (modern public domain version)
 * [Practical Common Lisp](http://www.gigamonkeys.com/book/) by Peter Seibel
 * [Common Lisp Recipes](http://weitz.de/cl-recipes/) by Edmund Weitz, published in 2016,
 * [Cliki](http://www.cliki.net/), Common Lisp's wiki


vindarel:
Hello, 
shall we put it on the clos page itself ? I think it's more appropriated.

Hexstream:
Violently disagree, but I'm tired of arguing about obvious stuff.

The CLOS MOP spec is almost like an official addendum to the CLHS.
closer-mop was the #3 Quicklisp library for all of 2017 according to Quicklisp download stats.

I don't care to say more about how profoundly important the CLOS MOP spec is, and why it shouldn't be buried in some page that very few people will visit instead of somewhere prominent along with other important resources.

vindarel:
Looks like you were tired and angry indeed.

I suggested to put this link on https://lispcookbook.github.io/cl-cookbook/clos.html, you might have not understood that.

Personally I don't see the point to add a link on the front page when the goal of the cookbook is to give recipes, with categories, and when the awesome-list already has those links.

I don't know why you speak about the CLUS.

Hexstream:
```
I suggested to put this link on https://lispcookbook.github.io/cl-cookbook/clos.html, you might have not understood that.
```

I did understand that you wanted to bury the link into a page that very few people would visit (https://lispcookbook.github.io/cl-cookbook/clos.html) instead of one that many, many more people would visit (https://lispcookbook.github.io/cl-cookbook/) (obviously a lot of people are going to just check the main page and then "bounce" for lack of time, possibly coming back later but not necessarily).

**There's already a link to the CLHS there, which has nothing to do with recipes.** As previously said:

```
The CLOS MOP spec is almost like an official addendum to the CLHS.
closer-mop was the #3 Quicklisp library for all of 2017 according to Quicklisp download stats.
```

**I really don't see why you'd want to put a link to the CLHS but not to "almost an official addendum to the CLHS" in this context, assuming your primary goal is helping users.**

vindarel:
> I really don't see why you'd want to put a link to the CLHS

because it's an old cookbook, with some cruft, because I didn't add it and I'm in favor of removing some links from this front page.

> bury the link into a big page that very few people would visit

I intend to show this link to people wanting to learn about CLOS… If people don't stop there, I don't see how this link about an important but advanced topic would help them.

> And where does it say this section is for recipe-style links only anyway??

it doesn't say, I don't say that either, my humble opinion open to *gentle discussion* is that we could  shorten it and put links on specific topics on their respective pages.

Hexstream:
Sorry, I thought this project was about helping users, that may have been a wrong impression.

vindarel:
errr, you're not helping us help users either O_o Did I say I was against your PR ? No. I was only trying to discuss to do better.

Hexstream:
I'm not helping you help users when I suggest to add a link to a relevant resource in an appropriate section already containing a link to a conceptually similar resource??

Well, I'm "not helping" in that once again someone prevents me from helping by preferring to argue about largely irrelevant details instead of instinctively helping users and opting to figure out the finer details later.

vindarel:
allright… thanks for the link.

Hexstream:
Better than nothing, thanks.

(I still don't understand how this isn't about on the same level of importance as the CLHS, but whatever.)

PR #165, opened 2018-03-01
More multidimensional arrays

bendudson:
Adds some BLAS calls, fixes a couple of typos, adds the outer-product function in `array-operations`
I've started a section on `Matlisp`, and slowly filling in as I figure it out. 

Is drip-feeding updates like this ok, or should I wait until it's all done (might be a very long wait...)?

### DIFF
--- a/arrays.md
+++ b/arrays.md
@@ -65,12 +65,13 @@ Some libraries available through
 [quicklisp](https://www.quicklisp.org/beta/) which take this approach 
 are: 
 
+* [matlisp](https://github.com/matlisp/matlisp), which is described in
+  sections below.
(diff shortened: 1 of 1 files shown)


vindarel:
Updates like this are very to me, I do similarly, otherwise the danger is to never send them…

I suggest we give you push access, shall we @pnathan ? (I can't do it) 
So than you can quickfix typos directly and send PRs with new content, so than followers can follow. There's also the CHANGELOG issue (152) for summaries.

vindarel:
Matlisp is now on Quicklisp: http://blog.quicklisp.org/2018/02/quicklisp-dist-update-for-february-2018.html

PR #166, opened 2018-03-05
More arrays

bendudson:
* Updated Matlisp intro, since now on Quicklisp
* Added section on converting between Matlisp tensors and CL arrays
* Added section on Matlisp element-wise operations
* Small update to nested-loop macro, using `loop` rather than `dotimes` and `nth`

### DIFF
--- a/arrays.md
+++ b/arrays.md
@@ -847,9 +847,9 @@ end point.
 ~~~
 
 ~~~lisp
-* (matlisp:linspace 1 (* 2 pi) 5)
+* (matlisp:linspace 0 (* 2 pi) 5)
 #<|<BLAS-MIXIN SIMPLE-DENSE-TENSOR: DOUBLE-FLOAT>| #(5)
- 1.000   2.321   3.642   4.962   6.283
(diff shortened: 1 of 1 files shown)


arrays.md
```diff
@@ -764,21 +762,15 @@ functions:
...
 ~~~lisp
 * (ql:quickload :matlisp)
 ~~~
 
-To avoid typing `matlisp:` in front of each symbol, you can use
+The nickname for `matlisp` is `m`, but to avoid typing `matlisp:` or
+`m:` in front of each symbol, you can use
 the package or for an interactive session run:
 ~~~lisp
 * (in-package :matlisp)
```
vindarel: what about also sometimes suggesting `use-package` ? (also re-showing a package definition using some libraries cannot hurt).

vindarel:
@pnathan, shall we give him push access ? (I don't have enough rights)

arrays.md
bendudson: Yes, good idea. I'll add something on using the packages.

PR #167, opened 2018-03-07
Fix broken link to Perl cookbook TOC

bendudson:
The current link is 404, but the TOC is listed on this page: http://shop.oreilly.com/product/9781565922433.do 

The link to http://www.oreilly.com/catalog/cookbook/ is currently forwarded to that page. Not sure which is more likely to be moved, but guessing that a product number may change.


vindarel:
nice catch ! (I thought we had an automatic checker… looks like we don't)

PR #170, opened 2018-04-03
Add section on circular lists to Data Structures

bendudson:
Creating using a function `circular!` or reader macro `#n=`.

Based on discussions/links here:
https://stackoverflow.com/questions/604988/self-referential-data-structures-in-lisp-scheme
https://stackoverflow.com/questions/16678371/circular-list-in-common-lisp
https://stackoverflow.com/questions/12649290/example-of-sharpsign-equal-sign-reader-macro

### DIFF
--- a/data-structures.md
+++ b/data-structures.md
@@ -78,6 +78,55 @@ or by calling quote:
 
 which is shorthand notation for the function call `(quote (1 2))`.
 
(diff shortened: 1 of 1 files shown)


vindarel:
Interesting, thanks!

PR #173, opened 2018-04-14
Add section on numbers

bendudson:
Started with the headings from the Perl cookbook and adapted a bit. Probably a lot more could be added, but I think it's in a reasonable state. Sorry for any errors/omissions.


vindarel:
Sooo nice :) You post on reddit ?

vindarel:
@pnathan Shall we give Ben push rights ? (I can't do it)

PR #174, opened 2018-05-06
Fix link to PCL files section

bendudson:
Link to Practical Common Lisp from the Files and Directories section
had a stray `s` at end of URL, and was not a hyperlink.


pnathan:
Thank you!

PR #187, opened 2018-09-28
CLOS - extensive rewrite

vindarel:


vindarel:
This wasn't too long to read over a week-end, was it ? :]

clos.md
```diff
@@ -2,1655 +2,1556 @@
...
-held in New York City in October 2003.
+* it is **dynamic**, making it a joy to work with in a Lisp REPL. For
+  example, changing a class definition will update the existing
+  objects, given certain rules which we have control upon.
+* it supports **multiple dispatch** and **multiple inheritance**,
+* it is different from most object systems in that class and method
+  definitions are not tied together,
+* it has excellent **introspection** capabilities,
+* it is provided by a **meta-object protocol**, which provides a
+  standard interface to the CLOS, and allows *in fine* to create new
```
bendudson: Not sure what's meant by "in fine" here

clos.md
```diff
@@ -2,1655 +2,1556 @@
...
+an ANSI standard ten years later.
 
-* an introduction to the 10% of CLOS which you need to get you
-through 90% of use cases;
-* a number of opinionated statements, for instance about the
-differences between CLOS and other object systems;
-* a brief look "under the hood" at how some of the more dynamic
-features of CLOS might be implemented;
-* some exercises.
+This page aims to give a good understanding of how to use CLOS, but little presents of the MOP.
```
bendudson: Suggest "but little presents of the MOP" to "but only a brief introduction to the MOP"

clos.md
```diff
@@ -2,1655 +2,1556 @@
...
+~~~lisp
+(defun make-person (name &key lisper)
+  (make-instance 'person :name name :lisper lisper))
+~~~
 
-CL-USER 8 > my-point
-#S(POINT X 3 Y -4 Z 12)
+This has the direct advantage that you can control the required
+arguments. You should now export the constructor from your package and
+not the class itself.
```
bendudson: Perhaps an example ASDF entry here would help explain how this is done, and emphasise the distinction between a constructor and the class.

vindarel: Not sure what to add, and I wouldn't add a package definition which exports symbols. I can remove the last sentence if it's more confusing than helping, as it seems to be.

vindarel:
@bendudson thank you.

So is it good enough (english-wise, content-wise) to be merged ?

bendudson:
I think it looks great, thanks very much. Definitely helped me understand CLOS better.

PR #190, opened 2018-11-23
package - Add rename-package

sheepduke:
Adds some introduction to `RENAME-PACKAGE`.

### DIFF
--- a/packages.md
+++ b/packages.md
@@ -39,23 +39,37 @@ Or you can do it with [`LOOP`][loop].
 
 ## Nickname Provided by Packages
 
-When defining a package, it is normal to give it a nickname for better user
-experience. For example:
+When defining a package, it is trivial to give it a nickname for better user
+experience. The following example is a snippet of `PROVE` package:
 
 ~~~lisp
-(defpackage :my-package-with-a-long-name
-  (:use :cl)
-  (:nicknames :my-package :my)
-  (:export :my-function
-           :my-another-function))
+(defpackage prove
+  (:nicknames :cl-test-more :test-more)
+  (:export :run
+           :is
+           :ok)
 ~~~
 
-Afterwards, a user may use nickname instead of the full name to refer to this
+Afterwards, a user may use nickname instead of the package name to refer to this
 package. For example:
 
 ~~~lisp
-(my-package:my-function)
-(my:my-another-function)
+(prove:run)
+(cl-test-more:is)
+(test-more:ok)
+~~~
+
+Please note that although Common Lisp allows defining multiple nicknames for
+one package, too many nicknames may bring maintenance complexity to the
+users. Thus the nicknames shall be meaningful and straightforward. For
+example:
+
+~~~lisp
+(defpackage #:iterate
+  (:nicknames #:iter))
+  
+(defpackage :cl-ppcre
+  (:nicknames :ppcre)
 ~~~
 
 ## Give a Package a Local Nickname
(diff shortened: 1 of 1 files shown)


packages.md
```diff
@@ -35,6 +35,26 @@ Or you can do it with [`LOOP`][loop].
   collect s)
 ~~~
 
+# Give package local nickname
```
vindarel: give *a* package *a* local nickname

packages.md
```diff
@@ -35,6 +35,26 @@ Or you can do it with [`LOOP`][loop].
...
+
+Sometimes it is handy to give imported a local name for saving some typing.
```
vindarel: IMO:

sometimes it is handy to give a local name to an imported package to save some typing.

packages.md
```diff
@@ -35,6 +35,26 @@ Or you can do it with [`LOOP`][loop].
...
+This can be achieved by using [`RENAME-PACKAGE`][rename-package]. For example:
+
+~~~lisp
+(asdf:load-system :cl-ppcre)
+
+(defpackage :mypackage
+  (:use :cl))
+(in-package :mypackage)
+
+(rename-package :cl-ppcre :re)
```
vindarel: Also a note about `:nickname` in package definitions ?

vindarel: maybe my remark is a better fit for systems.md. (do as you like)

packages.md
sheepduke: To be honest these contents was finished before a meeting so I was really in hurry.

I apologize for not double-checking before committing.

packages.md
sheepduke: I think might be better to clearify that "nickname" mentioned here is different with the one in `DEFPACKAGE`.

I will add it.

packages.md
sheepduke: Fixed. Please check.

packages.md
sheepduke: Fixed. Please check.

packages.md
sheepduke: Fixed. Please check.

packages.md
```diff
@@ -54,7 +78,58 @@ This can be achieved by using [`RENAME-PACKAGE`][rename-package]. For example:
 (re:scan "a" "abc")
 ~~~
 
+Please note that the original name (`:CL-PPCRE` in the example above) will
+**not** be available any more after renaming.
+
+## Package locks
+
+A package can be declared to be locked so that it cannot be modified by the
+user. Attempts to change its symbol table or redefine functions which its
```
vindarel: As a reader I ask myself: but why ? There's an answer in the SBCL doc:

> The package common-lisp and SBCL internal implementation packages are locked by default, including sb-ext. 

shall we (you :) ) add this explanation ?

vindarel:
Nice addition ! Another suggestion inside.

packages.md
```diff
@@ -35,6 +35,101 @@ Or you can do it with [`LOOP`][loop].
   collect s)
 ~~~
 
+# Package nickname
+
+## Nickname Provided by Packages
+
+When defining a package, it is normal to give it a nickname for better user
```
pnathan: I would note that that overuse of package aliases can wind up extremely confusing for future maintainers.

pnathan:
Just one comment, regarding maintainability concerns.

packages.md
sheepduke: As a man who has a bunch of nicknames, I would say "it depends".

If the nickname are meaningful and straightforward, I think it is OK. Also TBH I seldom see a package with more than one nickname.

But you are right. Giving too many nicknames to a package may lead to chaos especially when people use different nicknames simultaneously.

I will add some explanation of this. Thanks for your advice.

pnathan:
LGTM

sheepduke:
> Thanks, and pleasure to see a new contributor.

Welcome myself ;-)

PR #191, opened 2018-12-05
new section: debugging

vindarel:
This section is clearly lacking and needed for newcomers.

Any thoughts ? I think the content could be more thorough but is ok; I'd enjoy style & english improvements.


pnathan:
concern over the outlinks: too many links rot over time, can we pull the relevant information and cite the references?

vindarel:
You're right, I'll see what I can do. I don't think I need to pull new content, maybe screenshots, and I'll reference them in the end only.

vindarel:
Allright, I couldn't reproduce the nice interactive stories of the blog posts, that will be for later. I put the links at the bottom, with smaller edits.

PR #192, opened 2018-12-09
Use nickname instead of name in `RENAME-PACKAGE`

sheepduke:
After some consideration I think it is not good to recommend using `RENAME-PACKAGE` to directly modify the *name* of a package.

Instead, IMHO it is better to use `RENAME-PACKAGE` to add a new nickname.


vindarel:
cool, LGTM 

(I'll keep an eye to merge if needed)

PR #193, opened 2018-12-13
editors: add Slimv_box

vindarel:
dunno how convenient it is, but I'd like to list the options that are supposed to ease the installation of a development environment.


pnathan:
:+1:

PR #205, opened 2019-01-10
Series examples added.

paul-donnelly:
Added examples using Series for most of the iteration examples.


vindarel:
Veeery nice, thanks a lot for this contribution.

PR #219, opened 2019-03-07
[performance] Add performance tuning chapter

sheepduke:
#208

PR #221, opened 2019-03-13
performance: add assembly output and rephrase

vindarel:
"unused code are pruned" -> "the compiler was able to perform optimizations"

@sheepduke @bendudson thoughts ? Following our discussion on 
https://github.com/LispCookbook/cl-cookbook/pull/219/files#r264015818

### DIFF
--- a/performance.md
+++ b/performance.md
@@ -88,7 +88,7 @@ the declaration [`optimize`][optimize]. Each quality may be assigned a value
 from 0 to 3, with 0 being "totally unimportant" and 3 being "extremely
 important".
 
-The most significant qualities might be `safety` and `speed`. 
+The most significant qualities might be `safety` and `speed`.
 
 By default, Lisp considers code safety to be much more important than
 speed. But you may adjust the weight for more aggressive optimization.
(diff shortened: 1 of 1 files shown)


sheepduke:
I hid the assembly output because I was afraid the page would be too verbose.

Is it really necessary to show them in the content?

vindarel:
I guess it's allright: https://vindarel.github.io/cl-cookbook/performance.html#using-declare-expression
I like that it's easy to grasp that the assembler with type hints is really shorter. I suppose advanced readers (and we're in an advanced section) will want to read some assembler.

But the most important is this change

> "unused code are pruned" -> "the compiler was able to perform optimizations"

It must be in par with what you wanted to explain.

sheepduke:
> I guess it's allright: https://vindarel.github.io/cl-cookbook/performance.html#using-declare-expression
> I like that it's easy to grasp that the assembler with type hints is really shorter. I suppose advanced readers (and we're in an advanced section) will want to read some assembler.
> 
> But the most important is this change
> 
> > "unused code are pruned" -> "the compiler was able to perform optimizations"
> 
> It must be in par with what you wanted to explain.

Reasonable. Nice work!

PR #223, opened 2019-03-17
[performance] Enhance the content and fix some correctness issues

sheepduke:


sheepduke:
Also I would like to discuss the effect of `eval-when`.

According to this [Reddit comment](https://www.reddit.com/r/Common_Lisp/comments/b1qn4s/performance_tuning_and_tips_the_common_lisp/eio7pdf), it is meaningless to use `(eval-when (:compile-toplevel))`  together with `declaim` because `declaim` only works at top-level of a file to be compiled.

I am not really sure if it is true. I have never used `eval-when` myself.

Could anyone please help elaborate?

bendudson:
There is an interesting post here about `eval-when`: https://fare.livejournal.com/146698.html
I've also never used it: The one time I thought I needed it, it didn't behave as I expected, and it then turned out I didn't need it.

vindarel:
Can't say either.

btw @sheepduke I'll give you write access so than you can merge this kind of PR more easily. And you'll be able to push little quickfixes directly. cc @bendudson @pnathan

sheepduke:
> btw @sheepduke I'll give you write access so than you can merge this kind of PR more easily. And you'll be able to push little quickfixes directly. cc @bendudson @pnathan

Ha! Thanks for your trust. 

So minor fixes can be pushed directly?

What kind of changes should be considered big enough for a PR?

sheepduke:
> There is an interesting post here about eval-when: https://fare.livejournal.com/146698.html

The article does not encourage the use of `eval-when` in trivial cases.

After some consideration, I think it is better to remove `eval-when` related part for less misleading and chaos.

vindarel:
I added you as a collaborator, you should have received an invitation link.

> So minor fixes can be pushed directly?

Yeah I do it a bit. Ones that don't need a review or to be showed to followers (typos, fixing a code block and other work on the structure of the page). In case you want to polish your pages.

> What kind of changes should be considered big enough for a PR?

All, except when you're confident it isn't worth it. It's always better to make a PR so everybody is notified of a change. Leave it open for a few days to give a chance to reviewers, until one or two +1. If it's a small PR (like, "I learned this detail today and I wanted to add to the cookbook"), it's cool to make a PR to share with others, and I find ok to not wait for a review and merge after a couple of hours, while you're into it.

pnathan:
concur with @vindarel

PR #226, opened 2019-04-08
update Atom-slime -> SLIMA

vindarel:


sheepduke:
I do not use Atom so I cannot judge. But the expression LGTM.

pnathan:
@phoe strong agree

PR #227, opened 2019-04-12
Added an article about function declaration 

Sokolov57:
Mostly taken from http://www.lispworks.com/documentation/lcl50/aug/aug-36.html 
I hope this does not violate any copyright stuff, lispworks doesn't own the square function.
I just added some output for clarity.

### DIFF
--- a/performance.md
+++ b/performance.md
@@ -354,45 +354,37 @@ been compiled.
 
 
 ### Declaring function types
-
-A `ftype` declaration specifies the manner in which the returned value type of 
-a declared function depends on the argument types of the function. Whenever 
-the arguments to a declared function are of the indicated types, the result 
-of the function will also be of the indicated type. A function can have more 
-than one `ftype` declaration associated with it. One would use a`ftype` declaration 
-when you want to conditionally restrict the result of a function every time 
-the function is called. It has the following form:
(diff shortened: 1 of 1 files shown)


sheepduke:
I should thank you first because `ftype` part is in my TODO list but I was so busy recently...

vindarel:
They don't own the square function, but they own their text, and you copied sentences.

> All Rights Reserved. No part of this publication may be reproduced
http://www.lispworks.com/documentation/lcl50/aug/aug-3.html#HEADING3-0

:/

I think it's best anyways to digest and rewrite the content, to make it even clearer, and create another helpful resource, so people can read lispworks' and this. Thoughts ?

@pnathan ?

Sokolov57:
Yeah, ok, if nobody wishes to do it atm, I'll contribute to this rewriting part in an hour or so. I'm probably not very good at it, so pls check after me.

vindarel:
Thanks for this addition and for rewriting !

Sokolov57:
Well, the declamation -- I used it because it's outside of the scope, that's just how I am used to do this. But I have no authority what to use here, so..

vindarel:
Since @sheepduke approved, all good, thanks !

PR #235, opened 2019-05-21
Fix security vulnerabilities in Gemfile

eudoxia0:
This is not strictly necessary, but every week I get GitHub alerts over this, so I propose fixing it.


pnathan:
:+1:

vindarel:
Good move, thanks.

PR #239, opened 2019-06-15
Fix some broken links

ebzzry:


sheepduke:
LGTM.

vindarel:
Woah, thank you.

We should have an automated tool…

PR #242, opened 2019-06-17
Fix more broken links

ebzzry:


vindarel:
Thanks !

pnathan:
Thanks! As the web rots, these things happen, sadly enough. Your work is much appreciated.

PR #246, opened 2019-06-18
TOC now visible on small screens

digikar99:
Bootstrap helps in better fonts - therefore keeping it, otherwise, it's not used anywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ First option is to install [Jekyll][jekyll] and run `jekyll serve` in a folder w
 
 Another option is to use github gem provided with the repo. To do this do the following:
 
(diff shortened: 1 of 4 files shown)


vindarel:
Excellent change, thank you !

The blue is different (look at the two index pages) and I find your new blue too flashy. Thoughts ?

digikar99:
Okay, the new blue was due to bootstrap version change 3.3.7 -> 4. 

I had updated bootstrap to version 4, in the hopes of using their sidebar - but implemented it in native CSS as that was simpler, but left the bootstrap version updated. I have reverted this change, and the blue should be as before now.

PR #247, opened 2019-06-18
Fix broken links

ebzzry:


vindarel:
For the beautify extension: https://github.com/vale981/practical-cl-beautified/issues/1 Can you link to its Github instead, or wait a bit ? 

Also, feel free to refresh the windows section, it's quite old. emacs-ide also needs work.

ebzzry:
We can perform the merge now, because it’s uncertain whether that page will ever get back. If so, we can just update it later.

I’ll take a look at the Windows section more. It looks like it needs a lot of revisioning—some 3rd party Elisp libraries are now part of Emacs, at least with version 22.

Issue #105, opened 2017-07-28
Request: state of web development


vindarel:
For this I'd personally use https://stackoverflow.com/a/42838145/1506338 (and I'm very eager in writing it).

There would be examples, what's lacking, how to quickstart everything, references to libraries found in the Awesome list.

PuercoPop:
That SO post have several omissions. For example just form generators there is:

- [formlets](https://github.com/inaimathi/formlets)
- [cl-forms](https://github.com/mmontone/cl-forms)

Mariano also has [clavier](https://github.com/mmontone/clavier) for validation. The ORMs list fails to mention CL-SQL. Fukamachi has written [mito] recently

I also don't understand why one would mention wu-wei but not hunchentoot.

vindarel:
Thanks for the links ! They were hard to find. Just found also [1forms](https://github.com/tshatrov/1forms)

>  Django-like web (html) forms for Common Lisp,  It can be used with various Common Lisp web servers and web frameworks such as caveman2. 1forms uses CLOS and is extensible by design.

but not touched for 3 years.

Sure this post has omissions but 
- it is not meant to be exhaustive, this task is left to the Awesome CL list and the State of CL ecosystem 2015 (both links given in the intro)
- in my biased opinion (I'm the author there) it's the only place I know one can have an overview. 

It does mention Hunchentoot but puts emphasize on Clack.

> Maybe it would be a good idea to tackle this by sections first?

Ok. And shall we make up a list/our TOC first ?

can3p:
I think Hunchentoot is in much better place in terms of documentation. If clack is put in the cookbook it needs a really extensive tutorial there

PuercoPop:
The subsections/divisions could be

- Frameworks and Web Application Servers (Including websockets)
- Serialization/data storage (SQL and NoSQL?)
- Validation and form generation
- Misc (Job queues, oauth, jwt, etc)

vindarel:
+1 for the division. You didn't note javascript integration, but ok to do this last. Also we can suppose we'll have a real databases page, so we would speak here only about integrating a DB with a web framework.

Those are some points I'd like addressed (what I've been researching on this topic and what blocked me while trying Caveman and Lucerne).

First of all, what application shall  we write as an example throughout  the page ? (as in the web scraping page)
- use and extend the [Lucerne twitter clone example](http://eudoxia.me/lucerne/docs/example--a-twitter-clone.html) ?
[...]

PR #250, opened 2019-06-28
Web development

vindarel:
for #105 

I have a mixed feeling with this new page. It is needed, it has useful sections (storing passwords, building, deployments,…), it is an overview of a lot of stuff that otherwise need a lot of digging, but it isn't a proper tutorial like other pages. The reader can try the Hunchentoot examples, but not much more. It is also a half-baked Hunchentoot tutorial. I also don't really want to push for a framework, they have their pros and cons, so I wouldn't write a complete tutorial for one of them. That would be the role of another project.

I suggest we merge and complete missing sections (JS, form validation (there is no strong proponent), job queues,…).

https://vindarel.github.io/cl-cookbook/web.html

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -2,6 +2,9 @@ title: Common Lisp Cookbook
 description: A collection of examples of using Common Lisp
 baseurl: "/cl-cookbook"
 
+# https://stackoverflow.com/questions/39057405/unable-to-build-a-jekyll-site-invalid-date
+exclude: [vendor]
+
 markdown: kramdown
 sass:
   style: :compressed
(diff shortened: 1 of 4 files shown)


pnathan:
:eyes:  I'll stare at it later today when I need a break

pnathan:
> Hunchentoot and Clack are two web servers that you’ll often hear about.

Well, Clack isn't a web server, it's an application framework.

pnathan:
I like it, actually. I think it gives a good start.

vindarel:
> Well, Clack isn't a web server, it's an application framework.

fixed.

Thank you, I'll merge and iterate on it then !

PR #254, opened 2019-07-30
add duckduckgo search box

### LINKED ISSUE #252: search/index
it would be great if the site had a good search. it kind of does have an index now but im thinking of a real index where you hit a for all the topics that begin with a etc.

but search should really be pushed as a feature. i want to look something up either i need to use google(which is ok) or i need to wade through a bunch of stuff to find what im looking for.

vindarel:
Good and pretty enough ? https://vindarel.github.io/cl-cookbook/

fixes #252 (it's a simple search, not an index)

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -33,26 +33,44 @@
       <a href="{{ site.baseurl }}/">
         <img id="logo" src="{{ site.baseurl }}/assets/cl-logo-blue.png"/>
       </a>
+
+      <form onsubmit="duckSearch()" action="javascript:void(0)">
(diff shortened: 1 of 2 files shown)


pnathan:
:+1:

 this change really quacks

Regards,
Paul

markboston34:
As it was it was pretty hard to find stuff. An index would be good just to browse around if you didn’t know exactly what you were searching for.

vindarel:
I see. Do you have an idea/a tool in mind on how to build such an index ? 

Or, would you like the table of contents on the front page to be more precise ? (feel free to re-open your issue)

Issue #243, opened 2019-06-17
Examples on `iterate`


digikar99:
While exploring [The Iterate Manual](https://common-lisp.net/project/iterate/index.html), I created [this page on iterate](https://digikar99.github.io/cl-iterate-docs/), and shared it over [reddit](https://www.reddit.com/r/Common_Lisp/comments/bzw7l2/some_more_documentation_for_iterate_i_came_across/). [u/dzecniv](https://www.reddit.com/user/dzecniv/) suggested may be add it over here - I agree that I'd like to find all the Common Lisp content at one page. So, a request for addition of those examples over here.

I also see that [this issue is still open](https://github.com/LispCookbook/cl-cookbook/issues/168).

vindarel:
Hey, that would be an awesome addition. You can create a new page, and don't hesitate to add examples to the iteration one.

vindarel:
Hey, just discovered this: https://sites.google.com/site/sabraonthehill/loop-v-iter (this website has awesome articles)

arademaker:
Nice blog posts, it worth cite them but it would be good to keep the material self-contained.

PR #259, opened 2019-08-23
added page for iterate library

digikar99:
#243


iterate.md
```diff
@@ -0,0 +1,688 @@
+---
+title: iterate
+---
+
+`iterate` is a powerful iteration facility for Common Lisp, and a alternative to `loop`. As opposed to `loop`, `iterate` is more lispy, extensible, besides being more powerful.
```
vindarel: aN alternative

iterate.md
```diff
@@ -0,0 +1,688 @@
...
+## Documentation
+
+### accumulate
+`accumulate `*`expr`*` by `*`func`*` &optional initial-value `*`init-val`*` into `*`var`*
+
+This is a general-purpose accumulation clause. func should be a function of two arguments, the value of expr and the value accumulated so far in the iteration, and it should return the updated value. If no initial value is supplied, nil is used.
+
+
+```lisp
+CL-USER> (iter (for i in '(1 2 3))
```
vindarel: We don't have a style guide, but I'm for removing the `CL-USER>`. Yes ?

index.md
```diff
@@ -20,6 +20,7 @@ title: Home
 * [Strings](strings.html)
 * [Numbers](numbers.html)
 * [Loops, iteration, mapping](iteration.html)
++ [The Iterate Library](iterate.html)
```
vindarel: more logical to put it has a subsection ? (with an indentation level)

iterate.md
```diff
@@ -0,0 +1,688 @@
+---
+title: iterate
+---
+
+`iterate` is a powerful iteration facility for Common Lisp, and a alternative to `loop`. As opposed to `loop`, `iterate` is more lispy, extensible, besides being more powerful.
+
+See [The Iterate Manual] for a more detailed explanation of the various facilities. Also check [Other Resources on Iterate](#other-resources-on-iterate).
+
+## Highlights
```
vindarel: I find a bit rough to see high-level highlights straight away, without examples beforehand. Maybe, do you mean a reader comes here after having read the examples of the "iteration" page ?

digikar99:
Hmm... Thanks for reviewing.

> So, this is a reference rather than a recipe.
> 
> * you can reference it from the iteration page
> * did you check if you can expand or add examples there ?

Yup, I'll agree that it is more of a reference than a recipe - a recipe would be more like illustrating what "iteration really means"; for which, unfortunately, I do not have the time. So, should I close this PR, and push just the additional line pointing to the iteration page in the reference section?

vindarel:
If you want to add a link on the introduction of the iteration page to your hosted iterate reference, that's already good. That will already help readers find better explanations on iterate.

My main concern with this page as it is is:

- why have a proper page for iterate, why not include snippets in the `iteration` page? I think enhancing it would be great.
- its scope, being a reference, with high level explanations at the beginning instead of code snippets, is different than the rest of the Cookbook.

We can discuss :)

digikar99:
So, I have added the page in the [Reference Section](https://digikar99.github.io/cl-cookbook/iteration.html#credit-and-references), and added a few examples on the [iteration](https://digikar99.github.io/cl-cookbook/iteration.html) page itself.

vindarel:
This, I find great. We can appreciate the value of iterate at a glance. 

(Its feature to learn about clauses is particularly a pleasant surprise)

PR #274, opened 2019-09-18
walk directories

vindarel:


vindarel:
got feedback on lisp-tips: https://github.com/lisp-tips/lisp-tips/issues/21

PR #275, opened 2019-09-18
numbers: add literals for hex and other radixes

sheepduke:
I found this piece missing from the cookbook.

And it is really handy when writing some low-level code...


sheepduke:
Summon the grand @vindarel

Issue #104, opened 2017-07-28
Request: a page explaining and showing types support in CL


vindarel:
As someone curious about CL, I heard that one can compile a form and be notified at compile time of type errors. So CL has types ? How that ? How do they compare to Python or to Haskell ? Because we are in a cookbook I'd like many code samples.

As a practicing learner, I saw `declare` and sometimes `the`. I'd like a quick reference. I'd like to see how Slime will warn me, and if and when it is useful.

- declare CLHS: http://www.lispworks.com/documentation/lw51/LWRM/html/lwref-75.htm

a first search:
- great article: http://blog.30dor.com/2014/03/21/performance-and-types-in-lisp/
- two or three nice little examples: https://news.ycombinator.com/item?id=1154069
- CL the language: https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node103.html#SECTION001300000000000000000
- https://softwareengineering.stackexchange.com/questions/187316/types-in-lisp-and-scheme
- https://www.common-lisp.net/project/cmucl/doc/cmu-user/compiler-hint.html ?
- paper of the python CMU compiler with examples: http://sci-hub.cc/10.1145/141478.141558

vindarel:
Reference this:

https://medium.com/@MartinCracauer/static-type-checking-in-the-programmable-programming-language-lisp-79bb79eb068a

```
(defun meh (p1)
  (declare (fixnum p1))
  (+ p1 3))
(defun meh2 (p1)
  (declare (fixnum p1))
  (+ p1 "8"))

Gets you this:

2 compiler notes:typecheck-demo.lisp:7:3:

  note: deleting unreachable code

  warning:

    Constant "8" conflicts with its asserted type NUMBER.

    See also:

      SBCL Manual, Handling of Types [:node]Compilation failed.
```

Ambrevar:
And this: https://alhassy.github.io/TypedLisp/
By the way, this issue title mentions SLIME, but I think this is unrelated, this comes from compiler support.

SBCL might have the most extensive support for compile-time type checking, but it seems that CCL and ECL at least support some of it too.

It would be nice to highlight that the lack of static typing in Lisp is mostly a misconception.  And it would be even nicer to make those features more obvious to beginners :)

The correct term here seems to be not "static typing" but "gradual typing".

Finally, none of the examples show that you can type the return value of a function as well:

```lisp
(declaim (ftype (function (fixnum) fixnum) add))
(defun add (n)
[...]

PR #280, opened 2019-10-17
Compile-time type checking

vindarel:
for https://github.com/LispCookbook/cl-cookbook/issues/104

### DIFF
--- a/type.md
+++ b/type.md
@@ -324,7 +324,7 @@ We use again the `declaim` macro, with `ftype (function …)` instead of just `t
 
 With this we get nice type warnings at compile time.
 
-If we change the function to erronously return a string instead of a
+If we change the function to erroneously return a string instead of a
 fixnum, we get a warning:
 
 ~~~lisp
--- a/type.md
+++ b/type.md
@@ -269,6 +269,7 @@ thorough type checking.
 Let's recall first that Lisp already warns about simple type
 warnings. The following function wrongly wants to concatenate a string
 and a number. When we compile it, we get a type warning.
+
 ~~~lisp
 (defconstant +foo+ 3)
 (defun bar ()
--- a/type.md
+++ b/type.md
@@ -19,10 +19,10 @@ background.
 For example:
 
 ~~~lisp
-* (defvar *var* 1234)
+(defvar *var* 1234)
 *VAR*
 
-* (type-of *var*)
+(type-of *var*)
 (INTEGER 0 4611686018427387903)
 ~~~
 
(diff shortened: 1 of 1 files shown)


type.md
```diff
@@ -2,8 +2,11 @@
 title: Type System
 ---
 
-Common Lisp has a complete and flexible type system and corresponding tools to
-inspect, check and manipulate types.
+Common Lisp has a complete and flexible type system and corresponding
+tools to inspect, check and manipulate types. It allows to create
```
pnathan: It allows creating. Gerund form, not the infinitive.

type.md
```diff
@@ -2,8 +2,11 @@
...
+custom types, to add type declarations to variables and functions and
```
pnathan: `*adding`

type.md
```diff
@@ -250,13 +253,134 @@ The value of ARG is "Hello", which is not of type NUMBER.
...
+~~~lisp
+(setf *name* :me)
+Value of :ME in (THE STRING :ME) is :ME, not a STRING.
+   [Condition of type SIMPLE-TYPE-ERROR]
+~~~
+
+We can do the same with our custom types. Let's quickly declare the type `list-of-strings`:
+
+~~~lisp
+(defun list-of-strings-p (list)
```
pnathan: reflecting on this, this should be distinguished as a run time custom type checker, as opposed to a static type checker.  Since this code has to run every time the type check occurs.

type.md
```diff
@@ -250,13 +253,134 @@ The value of ARG is "Hello", which is not of type NUMBER.
...
+;     (VALUES STRING &OPTIONAL),
+;   conflicting with its asserted type
+;     FIXNUM.
+~~~
+
+This all happens indeed *at compile time*, either in the REPL,
+either with a simple `C-c C-c` in Slime, or when we `load` a file.
+
+
+After this, no need to recall that lisp already warns about simple
```
pnathan: I'm not sure what this question means rhetorically in the context.

type.md
vindarel: thanks!

type.md
vindarel: indeed. I kept the idea, rephrased and moved it at the beginning of the section.

type.md
vindarel: mmh suggestion on how to phrase this welcome.

vindarel: this code has to run every time the type check occurs, but it doesn't always occur at run time. When we use the declaim type declarations, it happens when we compile a function (thus acting as a static type checker).

vindarel:
up? And anything we can add on the topic?

vindarel:
Something to do: test with other implementations.

PR #284, opened 2019-11-07
getting started: CLISP repl is actually handy (it has completion)

vindarel:
It is yet another piece of information and more choice, but it solves an SBCL pain point. And since we still see people starting with CLISP, we might as well mention it once!

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -123,10 +123,15 @@ and:
 But we'll setup our editor to offer a better experience instead of
 working in this REPL. See [editor-support](editor-support.html).
 
-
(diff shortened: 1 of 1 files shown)


getting-started.md
```diff
@@ -111,7 +111,7 @@ distribution for more information.
 user@debian:~$
 ```
 
-You can slightly enhance the REPL (the arrow keys do not work,
+You can slightly enhance the SBCL REPL (the arrow keys do not work,
```
phoe: Not just SBCL repl. CCL, ECL, ABCL work too.

vindarel: right, thanks.

PR #286, opened 2019-11-13
Add information to on implicit mapping for Series.

paul-donnelly:


vindarel:
Thank you very much!

This apparently wasn't documented anywhere (https://www.reddit.com/r/lisp/comments/dt700a/so_series_apparently_supports_implicit_mapping/), now it is!

PR #287, opened 2019-11-13
`declare` is not a macro!, it is a symbol

nsrahmad:
`declaim` is a macro, and `proclaim` is a function. Also, updates the `declaim` link to go directly to CLHS without going to `xach.com` first.


vindarel:
Thank you.

PR #288, opened 2019-11-13
local-time: how to format time strings

vindarel:


vindarel:
Then I tried `cl-l10n` to print with the locale and the readme example didn't work: https://gitlab.common-lisp.net/cl-l10n/cl-l10n/issues/1

PR #300, opened 2020-02-12
updated expired CL standard link

alxnorden:
cvberry site for the CL Standard is link-rotted and now an aggressive link-farm. Updated to archive.org link for now, will have to review the copyright status on the PDF ( questionable license, as far as I understand)
Posting an update to an existing issue dealing with link-rot.

### DIFF
--- a/data-structures.md
+++ b/data-structures.md
@@ -10,7 +10,7 @@ also have many more details:
 - [Practical CL](http://gigamonkeys.com/book/they-called-it-lisp-for-a-reason-list-processing.html), by Peter Seibel
 - [CL Recipes](http://weitz.de/cl-recipes/), by E. Weitz, full of explanations and tips,
 - the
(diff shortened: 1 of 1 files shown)


vindarel:
Thanks! 

maybe a direct pdf download? https://gitlab.com/vancan1ty/clstandard_build/-/raw/master/cl-ansi-standard-draft-w-sidebar.pdf?inline=false

alxnorden:
We don't know what the status of the original author is.

With an archive.org link, we respectfully let people see some background of their work, and possibly see an update to the site if the author resurrects it (archive.org has a top bar that shows a more recent update)

With gitlab, if they change someting to the url (either the author changes something, or gitlab), a link to a (sub-part) of a repository is more efficient than a direct link, in my opinion. 

You decide.

alxnorden:
Got a reply from the author, he does not have a preference, so I suggest keeping link to (new) website and direct PDF download?

vindarel:
Great finds, thanks!

PR #304, opened 2020-02-21
Fix links

ebzzry:


web-scraping.md
```diff
@@ -148,7 +148,7 @@ reachable.
 
 External resources:
 
-- [CSS selectors by example](https://codingsec.net/2016/12/select-specific-text-css-using-selectors/)
+- [CSS selectors](https://www.w3schools.com/css/css_selectors.asp)
```
pnathan: https://developer.mozilla.org/en-US/docs/Glossary/CSS_Selector


let's use Mozilla's reference here.

ebzzry: Fixed!

ebzzry:
Thank you!

PR #334, opened 2020-06-19
Package local nicknames

bpanthi977:
Added description of `DEFPACKAGE` `:LOCAL-NICKNAMES` in packages.md

### DIFF
--- a/packages.md
+++ b/packages.md
@@ -117,6 +117,22 @@ above) will **not** be available any more after renaming. For example:
 This might cause problems if you want to load another package that relies on
 the modified package.
 
+### Package Local Nicknames (PLN)
+Many implementations (SBCL, CCL, ECL, Clasp, ABCL, ACL, LispWorks >= 7.2) support Package Local Nicknames (PLN). 
+
+~~~lisp 
+(defpackage :mypackage 
+  (:use :cl)
+  (:local-nicknames (:nickname :original-package-name)))
+  
+(in-package :mypackage) 
+
+;; You can use :nickname instead of :original-package-name 
+(nickname:some-function "a" "b")
+~~~
+
+Unlike `rename-package`, the effect of `PLN` is totally within `mypackage` i.e. the `nickname` won't work in other packages unless defined there too. So, you don't have to worry about unintended package name clash in other libraries.
+
 ## Package locks
 
 The package `common-lisp` and SBCL internal implementation packages are locked


packages.md
```diff
@@ -117,6 +117,22 @@ above) will **not** be available any more after renaming. For example:
 This might cause problems if you want to load another package that relies on
 the modified package.
 
+### Package Local Nicknames (PLN)
+Many implementations (SBCL, CCL, ECL, Clasp, ABCL, ACL, LispWorks >= 7.2) support Package Local Nicknames (PLN).
```
phoe, line 121: CLISP also supports them if built with the yet-unmerged commits from https://gitlab.com/gnu-clisp/clisp/-/merge_requests/3

vindarel:
Very welcome addition, thank you.

However our page introduces local nicknames by `rename-package` which is advised to NOT use… let's delete this section and only keep PLN? ping @sheepduke (who wrote about rename-package)

Also would you add a note on `:use`-ing a package being a bad practice?

phoe:
> However our page introduces local nicknames by `rename-package`

Wait a second - `RENAME-PACKAGE` is only capable of modifying **global** nicknames, not local nicknames.

Sure - I don't have much capability to focus on writing this though. Will you find a while to adapt the relevant parts from https://gist.github.com/phoe/2b63f33a2a4727a437403eceb7a6b4a3 that relate to how `:use` can be troublesome?

vindarel:
> Wait a second - RENAME-PACKAGE is only capable of modifying global nicknames, not local nicknames.

so we remove the section about it right?

I could, I was asking @bpanthi977 myself :)

phoe:
This commit says:

> Unlike `rename-package`, the effect of `PLN` is totally within `mypackage` i.e. the `nickname` won't work in other packages unless defined there too.

And this is correct; `rename-package` has global effects whereas PLNs have local effects.

However, the section "Give a Package a Local Nickname" should be named "Give a Package a **Global** Nickname" - the current version contains an error.

vindarel:
ACK.
But giving a global nickname will likely cause problems. I don't see when it would be useful compared to PLN. We should at least present PLN first, global one second.

phoe:
Exactly - that's the problem with global nicknames. They are rarely a good idea, since they pollute the common global namespace.

vindarel:
see https://github.com/LispCookbook/cl-cookbook/pull/335

sheepduke:
Original author here.

I admit that using `RENAME-PACKAGE` blindly is a bad practice. And writing this to the cookbook might make it worse.

I recall that I did not find a solid local nickname package at that time...
Maybe I was too stupid to find it ;-)

sheepduke:
Anyway, thanks for fixing this!

PR #335, opened 2020-06-19
packages: more basic, use bad practice

vindarel:


packages.md
```diff
@@ -72,66 +135,35 @@ example:
...
-(defpackage :mypackage 
-  (:use :cl)
-  (:local-nicknames (:nickname :original-package-name)))
-  
-(in-package :mypackage) 
+Prior to package-local-nicknames, users could define *global* ones
+with [`RENAME-PACKAGE`][rename-package]. However, the original name
+would **not** be available any more after renaming to **anyone**. If
+you renamed `cl-ppcre` to `re`, another Quicklisp library that uses
+`cl-ppcre` would fail. Use PLN.
```
bpanthi977, line 165: Saying that the 'original name would not be available' is a little misleading. As `rename-package` can be used to add global nicknames without removing original package name. (As given in original example)
```
;; Add nickname :RE to package :CL-PPCRE.
(rename-package :cl-ppcre :cl-ppcre '(re))
```
I suggest we direct readers to HyperSpec in that matter and warn about not using `rename-package`

```suggestion
Another facility exists for add nicknames to packages. The function [`RENAME-PACKAGE`](http://www.lispworks.com/documentation/HyperSpec/Body/f_rn_pkg.htm) can be used to replace the name and nicknames of a package. But it's use would mean that other libraries may not be able to access the package using the original name or nicknames. There is rarely any situation to use this. Use Package Local Nicknames instead.  
```

PR #339, opened 2020-07-02
getting-started.md:  correcting an oversight in the quicklisp installation instructions

various-and-sundry:
The command for installing quicklisp in a user defined directory was missing a final quotation mark around the file path.


vindarel:
Thank you!

PR #340, opened 2020-07-13
slime-load-system

mmontone:
I think slime-load-system is worth mentioning, as it allows the inspection of an ASDF system errors and warnings interactively (very handy).


vindarel:
thanks!

PR #342, opened 2020-07-13
SLIME Evaluation vs compilation: pragmatic reasons

mmontone:
After 15 or more of programming in Common Lisp, I've realized that compiling top-level forms is better than evaluating them, for the reasons I explain there (errors highlighting and line tracking). I would like to mention this so that others don't make my same mistake of using evaluation and missing what compilation provides.

Also, I think you use C-c C-c for compiling top-level forms, not C-c C-e (have a look at the diff)

Feel free to rephrase as you wish! My English is not the greatest.

### DIFF
--- a/emacs-ide.md
+++ b/emacs-ide.md
@@ -236,7 +236,7 @@ Compile a **region** by selecting the first 2 forms in test-all and
 running `M-x slime-compile-region`.
 
 Compile a **defun** by putting the cursor inside the "test-format"
-defun and pressing `C-c C-e`.
+defun and pressing `C-c C-c`.
 
 Compile the **sexp** before the point by putting the cursor after the
 closing paren of `(test-format)` and pressing `C-c C-e`.
@@ -246,6 +246,16 @@ To **evaluate** rather than compile:
 - evaluate a defun with `C-M-x`,
 - evaluate the sexp before the point with `C-x C-e`.
 See also other commands in the menu.
+
+---
+**EVALUATION VS COMPILATION**
+
+There are a couple of pragmatic differences when choosing between compiling or evaluating.
+In general, it is better to *compile* top-level forms, for two reasons:
+* Compiling a top-level form highlights warnings and errors in the editor, evaluation does not.
+* SLIME keeps track of line-numbers of compiled forms; when a top-level form is evaluated, the file line number information is lost.
+That's problematic for code navigation afterwards.
+---
 
 <a name="Slide-12"></a>
 


emacs-ide.md
```diff
@@ -236,7 +236,7 @@ Compile a **region** by selecting the first 2 forms in test-all and
 running `M-x slime-compile-region`.
 
 Compile a **defun** by putting the cursor inside the "test-format"
-defun and pressing `C-c C-e`.
+defun and pressing `C-c C-c`.
 
 Compile the **sexp** before the point by putting the cursor after the
 closing paren of `(test-format)` and pressing `C-c C-e`.
```
vindarel, line 242: Doesn't this command run eval instead of compile?

emacs-ide.md
```diff
@@ -247,6 +247,16 @@ To **evaluate** rather than compile:
...
+---
+**EVALUATION VS COMPILATION**
+
+There are a couple of pragmatic differences when choosing between compiling or evaluating.
+In general, it is better to *compile* top-level forms, for two reasons:
+* Compiling a top-level form highlights warnings and errors in the editor, evaluation does not.
+* SLIME keeps track of line-numbers of compiled forms; when a top-level form is evaluated, the file line number information is lost.
+That's problematic for code navigation afterwards.
+---
+
```
vindarel, line 259: An example when sometimes you prefer to eval?

emacs-ide.md
mmontone: C-c C-c runs compile. C-c C-e runs evaluate.

vindarel: > Compile […] pressing `C-c C-e`.
> ^^^                                      ^^

then it should say "Evaluate"?

emacs-ide.md
mmontone: To compile any defun or class. Have you tried this yourself? Perhaps it is better if you try it and see what I mean.

Say you have this function in your file:

```lisp
(defun foo (x))
```
Go to a function and evaluate it instead of compiling it (C-c C-e). You won't see any warnings highlighted in source, only warnings in repl.

After that, try navigating to FOO using M-. . You want be able to (slime-edit-definition-cont: Error: DEFINITION-SOURCE of function FOO did not contain meaningful information.)

Now compile it instead (C-c C-c). You will see the warning highlighted in editor.  And try navigate to it with M-. . Success!

vindarel: OK. And in the contrary, are there some situations where you don't compile, but prefer to eval? (that would be to illustrate further what your wrote, suggesting that eval has its use)

emacs-ide.md
mmontone: Also, I think evaluation commands are wrong there?

C-c C-e, prompts for evaluation.

But the evaluation described there I think corresponds to C-x C-e (slime-eval-last-expression).

https://common-lisp.net/project/slime/doc/html/Evaluation.html

mmontone: I hope I'm not getting confused here ...

mmontone: > > Compile […] pressing `C-c C-e`.
> > ^^^                                      ^^
> 
> then it should say "Evaluate"?

I thought compilation was being described there: "Compile a **defun** by putting the cursor inside the "test-format""

So I replaced eval command by compile command.

vindarel: I'm very sorry, my quote was misleading. Your change is right, I was wondering about the next sentence!

Not a big deal, you can leave your change and end the discussion, it's fine!

emacs-ide.md
mmontone: Use eval to observe results from non top-level forms.

Say you have this function:

```
(defun foo ()
  (let ((f (open "/home/marian/test.lisp")))
    ...))
```
Go to the end of the OPEN expression and evaluate it (C-x C-e), to observe the result:

```
=> #<SB-SYS:FD-STREAM for "file /mnt/e6b00b8f-9dad-4bf4-bd40-34b1e6d31f0a/home/marian/test.lisp" {1003AAAB53}>
```

vindarel: oh, indeed. I think that would make a great addition :) 

(I don't use that much myself, I'll try)

emacs-ide.md
mmontone: Ok. I think we may want to revise the evaluation commands after. I think they should be C-x C-e instead of C-c C-e most of the time.

emacs-ide.md
mmontone: I think that Evaluation appearing before Compilation in the SLIME manual is unfortunate.

Because when you are learning Lisp and SLIME, you learn to evaluate before learning to compile, and may never see the difference.

That's why I'm trying to highlighting the difference in the cookbook.

mmontone: These things are easy to miss just reading a manual. Perhaps it is better to highlight them in a video. But anyway ...

PR #349, opened 2020-09-07
hash-tables: pretty printing * 2

vindarel:


data-structures.md
```diff
@@ -1139,6 +1139,34 @@ CL-USER> (hash-table-count *my-hash*)
 
 ### Printing a hash table readably
 
+**With print-object**
+
+If you consider adding the function yourself, you can use `print-object`:
```
phoe, line 1144: Note that this is undefined behavior, since the standard doesn't permit the user to `(defmethod print-object ((object hash-table) stream) ...)`. We are essentially redefining standard Common Lisp behavior of printing hash tables.

A somewhat more annoying but portable way is https://github.com/phoe/phoe-toolbox/blob/master/phoe-toolbox.lisp#L464-L471

vindarel: damn. So I'll remove the print-object method. Yours is already in the recipe.

phoe: If you take a look at [what RUTILS does](https://github.com/vseloved/rutils/blob/eedd9b2b0afc90c98a2a2756f35b000743aa5885/core/hash-table.lisp#L192-L193), it does the same: it modifies the standard method for printing hash tables, too. You might therefore want to add a mention that this behavior is not portable.

vindarel: thanks. Do we know under which implementations overriding print-object works? We could use a `#+` tag, couldn't we? at least `#+sbcl`

vindarel: I added the mention.

PR #350, opened 2020-09-10
Codebase navigation tip

mmontone:
Use slime-edit-definition with a prefix argument in order to autocomplete and then navigate to any loaded definition.

Feel free to rephrase or discard if you don't think this tip is valuable, but I found it very handy now that I've discovered it.

### DIFF
--- a/emacs-ide.md
+++ b/emacs-ide.md
@@ -323,6 +323,12 @@ See also helm-imenu and [imenu-anywhere](https://github.com/vspinu/imenu-anywher
 Put the cursor on any symbol and press `M-.` to go to its
 definition. Press `M-,` to come back.
 
+---
+**CODEBASE NAVIGATION TIP**
+
+Use `M-- M-.` (slime-edit-definition with a prefix argument) in order to autocomplete and then navigate to any loaded definition.
+---
+
 #### Crossreferencing: find who's calling, referencing, setting a symbol
 
 Slime has a nice cross referencing facility, for example, you can see
--- a/emacs-ide.md
+++ b/emacs-ide.md
@@ -320,13 +320,17 @@ See also helm-imenu and [imenu-anywhere](https://github.com/vspinu/imenu-anywher
 
 #### Go to definition
 
-Put the cursor on any symbol and press `M-.` to go to its
+Put the cursor on any symbol and press `M-.` (`slime-edit-definition`) to go to its
 definition. Press `M-,` to come back.
 
 ---
 **CODEBASE NAVIGATION TIP**
 
-Use `M-- M-.` (slime-edit-definition with a prefix argument) in order to autocomplete and then navigate to any loaded definition.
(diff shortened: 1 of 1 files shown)


vindarel:
Is it supposed to complete a symbol name? If I try with `(forma`, I get "No known definition for: forma".

And when the symbol is complete, it is the same behavior as `M-.`?

thanks

mmontone:
Yes, it is the equivalent of M-.,but you don't need to enter anything on the repl or any file. Completes directly from Emacs command bar. It is the equivalent of M-. when cursor is on an empty area.

Perhaps the description of M-. is enough there in the Cookbook page, but I've found that I can navigate my whole Lisp project just thinking of the objects defined in them (classes, methods, functions and variables), and don't need to think of the files where they are defined at all.

Feel free to do whatever you want with this pull request :)

Cheers!

Mariano

mmontone:
And if you add SLIME [fuzzy completion](https://common-lisp.net/project/slime/doc/html/Fuzzy-Completion.html) you get super powerful navigation.

mmontone:
This is how navigation feels like: https://youtu.be/ZAEt73JHup8

vindarel:
Thank you! So I can reproduce when in the REPL, but not in a lisp file. I can't get the Emacs command bar when in a file :/ 

Is it possible for me to edit this PR?

mmontone:
>> Thank you! So I can reproduce when in the REPL, but not in a lisp file. I can't get the Emacs command bar when in a file :/

That's odd! Should work either with M-. , when you are on an empty text area. Or, as I explain, using M-- M-. (equivalent to slime-edit-definition with a [prefix argument](https://www.emacswiki.org/emacs/PrefixArgument)) wherever the cursor is.

See if you are getting the prefix argument right.

This is slime-edit-definition, that changes its behavior when prefix argument is present:

```elisp
(defun slime-edit-definition (&optional name where)
  "Lookup the definition of the name at point.
If there's no name at point, or a prefix argument is given, then the
function name is prompted."
  (interactive (list (or (and (not current-prefix-arg)
                              (slime-symbol-at-point))
                         (slime-read-symbol-name "Edit Definition of: "))))
  ;; The hooks might search for a name in a different manner, so don't
  ;; ask the user if it's missing before the hooks are run
  (or (run-hook-with-args-until-success 'slime-edit-definition-hooks
                                        name where)
      (slime-edit-definition-cont (slime-find-definitions name)
                                  name where)))
```

>> Is it possible for me to edit this PR?

Yes. Or write what you want here and I'll change it. Or just ignore it and add a tip yourself trying to explain this as well as you can. I'm not very eloquent in English :-)

Cheers!

vindarel:
I edited it, explained more and linked to your video :)

and silly me, I had lispy-mode enabled instead of the default lisp-mode.

mmontone (commented):
Good! Just merge.

PR #352, opened 2020-09-22
add FUNDING.yml

vindarel:
Hello,

So as I gather information about CL developers accepting donations (https://github.com/vindarel/lisp-maintainers/pull/1), I also try to push myself, since I am in a period where I don't have a fixed, nor big, income.

I added myself here (since I am the main contributor and I am still active) but actually we can list more than one contributor. The "♥ Sponsor" button that would appear can display github sponsors links for up to 4 people, and we can add up to 4 custom links.

So I would like to be sure that this PR doesn't cause a problem to anybody. 

And if this new button caused uncertainty and doubt in the future, the maintainers (including myself) can still revert this change.

Best,


pnathan:
:+1:

vindarel:
Then I merge, thanks!

(it's not too late to speak up)

PR #355, opened 2020-10-02
functions: referencing by name, single quote or sharpsign-quote?

vindarel:


functions.md
```diff
@@ -356,6 +356,48 @@ can be used on a list, for example from `&rest`:
 (apply #'+ '(1 2))
 ~~~
 
+### Referencing functions by name: single quote `'` or sharpsign-quote `#'`?
+
+In the example above, we used `#'`, but a single quote also works, and
+we can encounter it in the wild. Which one to use?
+
+It is generally safer to use `#'`, because it respects lexical scope.
```
phoe, line 364: It is generally safer to use `#'`, because it respects lexical scope. If we pass a symbol to `funcall`, what is called is always the function named by that symbol in the global environment, whereas using the `function` operator or its `#'` shorthand allows us to refer to local functions.

vindarel: ok, added ✔

PR #356, opened 2020-10-05
rm cl21 from the index. We can't recommend it (anymore…)

vindarel:
I admit we can't recommend CL21[1].

I'd still like a batteries-included CL package to appear though.

On a positive note, as a beginner, I would have many more friendly libraries at my disposal (than when I was looking at CL21): generic-cl, access, rutils (now with documentation), and many smaller ones.

[1] (even though it still works in my project©…)

### DIFF
--- a/cl21.md
+++ b/cl21.md
@@ -2,6 +2,12 @@
 title: Common Lisp for the 21st century
 ---
 
+CAUTION: We were excited by `CL21`, but we must admit that the project
+is staling and suffers from unresolved issues. We can't recommend it
+anymore. Hopefully, you'll now find many equivalent features in other
+libraries (`generic-cl`, `rutils`, `access` and many more).
+
+
 [CL21](http://cl21.org/) is an experimental project redesigning Common
 Lisp. It makes some common things that were tedious to do much easier
 and brings some new welcome features:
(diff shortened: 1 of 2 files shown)


pnathan:
Do you have more thoughts on why we can't recommend it?

phoe:
Some previous discussions:

* https://www.reddit.com/r/lisp/comments/6t6fqs/which_sugared_library_do_common_lispers_prefer/
* https://www.reddit.com/r/lisp/comments/6snw5d/questions_for_2017_common_lisp_experts/dlfi78d/

vindarel:
From the discussions, I noted one valid issue: https://github.com/cl21/cl21/issues/100

the project is dormant, the maintainer is not answering to PRs.

phoe:
It might be best to fork this project and maintain it there, since the maintainer is occupied with private matters - see https://github.com/quickdocs/quickdocs/issues/47

pnathan:
Works for me. I never cared for the library myself; but didn't really try and break out why or dig into it.

cl21.md
```diff
@@ -2,6 +2,12 @@
 title: Common Lisp for the 21st century
 ---
 
+CAUTION: We were excited by `CL21`, but we must admit that the project
+is staling and suffers from unresolved issues. We can't recommend it
+anymore. Hopefully, you'll now find many equivalent features in other
+libraries (`generic-cl`, `rutils`, `access` and many more).
+
+
 [CL21](http://cl21.org/) is an experimental project redesigning Common
```
pnathan, line 11: just remove cl21 altogether, don't mention it.

vindarel: Sure? In case there are external links pointing to it, it's friendly to not "break google".

I removed it from the index, so newcomers shouldn't see it.

PR #360, opened 2020-11-05
iteration: a word about GTWIWTG, similar to Series

vindarel:


phoe:
I'd note that it's GPLv3 and therefore under a copyleft license, unlike the majority of Common Lisp libraries out there.

vindarel:
yes. But mmh that may change, and we might not update it here and mislead readers. Maybe the mention on GitHub is obvious enough?

Or we write with precautions: "at the time of writing, the license is…"

PR #364, opened 2020-12-07
Fix format directive

as-edrisy:
The ~y directive gives an error
error in FORMAT: Unknown directive (character: LATIN_CAPITAL_LETTER_Y)


phoe (approved):
An obvious typo; thanks for catching that!

PR #366, opened 2020-12-17
EPUB and PDF versions

vindarel:
Hello,

I am happy to present an EPUB and a PDF version of the Cookbook.

I didn't commit them yet. Get theme here:

EPUB: https://framadrop.org/lufi/r/CaxEMv8MwC#QdhkCEe1M307X2DC4Bu3/N99bJJBqLd4+DwxP0/0ta0=

PDF: https://framadrop.org/lufi/r/dL_j0JDkUM#cCA6QitRHN3MKpRhR56XaviHZfAXcqc7NQv+y69L/2I=

I edited the content for the EPUB first during the last month, and I created the PDF from it. So the PDF isn't first-class, but it's OK. For a better PDF we'd have to write LaTeX and use other build rules, I believe.

We ship the PDF too, right? 

The front page looks like: https://vindarel.github.io/cl-cookbook/ There will be a direct download link, but also a green button to ["pay what I want"](https://ko-fi.com/s/01fee22a32), like on Leanpub where, if the author chooses so, the e-book is available both for free and for a custom amount. 

There is an announcement at the top of all pages.

I added a foreword and as an appendix the list of all contributors so far, without forgetting to thank the people who contribute by reviewing pull requests but whose name doesn't appear in the git history. 

Time to review! (especially the wording of the new pages :) )

ps: what would be the best EPUB file name?


vindarel:
Oh, is the "Using the Win32 API" worth keeping? (it uses Lispworks for Windows version 4.2.7).

vindarel:
Style issue :/ We're supposed to see 




as it correctly shows on my gh-pages

PR #369, opened 2021-01-05
fix emacs-ide page

vindarel:


vindarel:
How it renders: https://vindarel.github.io/cl-cookbook/emacs-ide.html

I'd like to add a section on how to setup Slime from scratch, but later, this PR is to close the related EPUB issue.

PR #372, opened 2021-01-26
daewok Docker images have moved to clfoundation

daewok:
Additionally, revamp the CI section that used them to use my (potentially overly) opinionated recommendation.


daewok:
I'm happy to roll back some of the Gitlab CI changes, but I think what this PR changes it to is a better recommendation overall. The analog to `daewok/lisp-devel` is `clfoundation/cl-devel`. It has all the implementations, Quicklisp is preinstalled, etc. But it is massive and gives ~zero control over the versions used.

By using the [implementation specific images](https://common-lisp.net/project/cl-docker-images/images/implementation-specific.html), you can specify the versions of each implementation to test against. You can even test against nightly versions (but I didn't want to advertise that in the cookbook in case it ends up being unsustainable to build them). Additionally, by installing Quicklisp as part of CI, you have control over which dist version to use).

vindarel:
Thank you!

PR #383, opened 2021-06-05
Fix typos

jgarte:


clos.md
```diff
@@ -406,7 +406,7 @@ The [closer-mop](https://github.com/pcostanza/closer-mop) library is *the*
 portable way to do CLOS/MOP operations.
 
 
-A subclass inherits all of its parents slots, and it can override any
```
pnathan, line 409: parents'

pnathan (changes requested):

clos.md
jgarte: Thanks for catching that. I pushed again with your fix.

vindarel:
Thanks +1

PR #385, opened 2021-06-17
Don't wrap SWANK:CREATE-SERVER inside BT:MAKE-THREAD

iamFIREcracker:
In general, a call to SWANK:CREATE-SERVER is not blocking (i.e. it
creates a thread on its own), hence wrapping it inside BT:MAKE-THREAD is
kind of redundant.  The only situations for which a call to
BT:MAKE-THREAD would be required are as if:

- the underlying implementation does not support multi-threading, in
which case, a call to BT:MAKE-THREAD would fail anyway
- the user is explicitly forcing a _blocking_ communication style (i.e.
`(setf swank:*communication-style* nil)`), but since we, as the users,
are not enforcing that, then we should be good without that extra call

While at it, I also:

- made QL:QUICKLOAD calls consistent by specifying packages as strings
(previously it was a mix of strings and keywords)
- wrapped the infinite loop inside BT:MAKE-THREAD, that way the user can
interact with the image while it's running (i.e. call BT:ALL-THREADS to
confirm the Swank server is up and running)

### DIFF
--- a/debugging.md
+++ b/debugging.md
@@ -397,7 +397,7 @@ If needed, import the dependencies first:
   (swank:create-server :port 4006)
   (format t "we are past go!~%")
   (bt:make-thread (lambda ()
-                    (loop while t do
+                    (loop repeat 5 do
                           (sleep 5)
                           (dostuff)
                           (incf *counter*)))
--- a/debugging.md
+++ b/debugging.md
@@ -376,7 +376,7 @@ To test this, let's define a function that prints forever.
 If needed, import the dependencies first:
 
 ~~~lisp
-(ql:quickload '(:swank :bordeaux-threads))
+(ql:quickload '("swank" "bordeaux-threads"))
 ~~~
 
 
@@ -394,14 +394,14 @@ If needed, import the dependencies first:
   (format t "hello world ~a!~%" *counter*))
 
 (defun runner ()
-  (bt:make-thread (lambda ()
-                    (swank:create-server :port 4006))
-                  :name "swank")
+  (swank:create-server :port 4006)
   (format t "we are past go!~%")
-  (loop while t do
-       (sleep 5)
-       (dostuff)
-       (incf *counter*)))
+  (bt:make-thread (lambda ()
+                    (loop while t do
+                          (sleep 5)
+                          (dostuff)
+                          (incf *counter*)))
+                  :name "do-stuff"))
 
 (runner)
 ~~~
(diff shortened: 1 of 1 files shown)


vindarel:
Thanks!

Could you add a snippet to kill the do-stuff thread? So a beginner doesn't have to search for it elsewhere.

phoe:
I'd object and instead add a comment that using `bt:destroy-thread` is a **bad** idea in the general case. Threads should instead finish on their own to do all the cleanup that they might need to do - one should never use `bt:destroy-thread` in production code. Destroying threads from outside is only feasible as a measure of last resort, as it can and eventually will leave the application in an unknown state.

It is better to `bt:interrupt-thread #'break` and select the bottommost `abort` restart in order to let the thread unwind completely. A programmatic version of this that works at least on SBCL would be `(bt:interrupt-thread (bt:current-thread) (lambda () (invoke-restart (alexandria:lastcar (remove-if-not (alexandria:curry #'eq 'abort) (compute-restarts) :key #'restart-name)))))`. A readable way of that magic one-liner would be:

```lisp
(defun abort-thread (thread)
  (flet ((thunk ()
           ;; We must call COMPUTE-RESTARTS in the dynamic context of the thread
           ;; that we want to abort, not of the thread that performs the
           ;; interruption.
           (let* ((restarts (compute-restarts))
                  ;; We assume that the last ABORT restart causes the whole
                  ;; thread to unwind and exit gracefully.
                  (abort-restarts (remove-if-not (alexandria:curry #'eq 'abort)
                                                 restarts :key #'restart-name))
                  (abort-thread-restart (alexandria:lastcar abort-restarts)))
             (invoke-restart abort-thread-restart))))
    ;; Interrupt the target thread with our thunk and force it to invoke the
    ;; ABORT restart, therefore initiating the unwind.
    (bt:interrupt-thread thread thunk)))
```

vindarel:
oops, thanks but that would be a bit much for our use case^^

Instead, use `dotimes` instead of an infinite loop?

phoe:
That would be good, too. (Maybe the above deserves its own section if it does not exist yet?)

vindarel:
indeed. On the Threads page (https://lispcookbook.github.io/cl-cookbook/process.html#joining-on-a-thread-destroying-a-thread), it is only mentioned that destroying a thread is not recommended

> for destroying a thread (not a recommended operation), we can use the bt:destroy-thread function.

phoe:
I assume that my snippet from above can be added there, yet. I'll create a separate issue for that.

pnathan:
> I'd object and instead add a comment that using bt:destroy-thread is a bad idea in the general case. Threads should instead finish on their own to do all the cleanup that they might need to do - one should never use bt:destroy-thread in production code. Destroying threads from outside is only feasible as a measure of last resort, as it can and eventually will leave the application in an unknown state.

concur

iamFIREcracker:
> Instead, use `dotimes` instead of an infinite loop?

Replaced `(loop while t ...` with `(loop repeat 5 ...`, let me know if there is anything else you want me to change.

PR #396, opened 2021-08-15
Update iteration.md

alx-a:
1. Fixed missing )
2. Updated broken resources URLs, when possible
3. Removed one dead link and pinged author with request for update 
4. Added author names to references, where available
5. Linked to Loop in the CL draft, as it appears the most verbose doc available on it? There is strong correlation with  the Steele book, but in my opinion they compliment each other. Discuss?
6.  [?] Am I using the correct / canonical reference for Steele?


vindarel:
Thank you!

Issue #399, opened 2021-08-21
ePub/Pdf not includes "Data Structures" chapter.


sigterm015:
Hello!

data-structures.md is missing on make-cookbook.lisp.

Thanks.

alx-a:
Weirdness with the Desktop client, i'll just give it a rest for now. Go with  a44b11a

vindarel:
ooooopsssi, thank you!

PR #400, opened 2021-08-22
Contents update

alx-a:
Address  #399


alx-a:
Weirdness with desktop client, ignore.

Issue #204, opened 2019-01-07
dead links ahoy


pnathan:
https://www.site24x7.com/link-checker.html

The internet rots under us, alas. We can but repair.

http://www.psg.com/~dlamkins/sl/cover.html should, probably, be amended to http://successful-lisp.blogspot.com/p/httpsdrive.html 

http://www.aiai.ed.ac.uk/~jeff/lisp/cl-pitfalls is now gone; it looks like `~jeff` is gone.

odanoburu:
> http://www.aiai.ed.ac.uk/~jeff/lisp/cl-pitfalls is now gone; it looks like ~jeff is gone.

wayback machine to the rescue! https://web.archive.org/web/20171007090745/http://www.aiai.ed.ac.uk/~jeff/lisp/cl-pitfalls

markboston34:
is this issue still looking for someone to help?

vindarel:
if you find an automated tool for that, that would be very nice ! (there must be many out there)

alxnorden:
Hope it's ok i'm upping here instead of creating a new issue. 
Just pulled a request on an updated link to the CL standard, the cvberry website appears to have become a link farm. More details in the request.
It appears link rot is a serious concern, that particular web squatter I fixed is full of aggressive JS.

I'm going to slowly review other external links in the cookbook. The more people do this, the better.

alx-a:
So I ran markdown-link-check, https://www.npmjs.com/package/markdown-link-check
and found some 404s. That markdown-link-check can be integrated into GH workflows https://github.com/gaurav-nelson/github-action-markdown-link-check
* Discuss if it's worth the trouble?
* Of course the broken Web of today means that many servers don't provide the courtesy of a 404 anymore, so manual checking and re-checking still applies.
* Since the links in question often need discussion, doing small individual PRs, hope it's ok, otherwise please let me now.

vindarel:
This anti-404 Github action would help :) Do we need it on every push though? Is there a solution to run it periodically? (I'm only thinking it would be wasted energy, I am optimizing prematurely…) Does a `[ci skip]` text in a message commit skips the CI with Github Actions? (that's a Gitlab CI thing)

edit: for a schedule:

```
schedule:
  # Run everyday at 9:00 AM (See https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07)
  - cron: "0 9 * * *"
```

alx-a:
[...]

PR #401, opened 2021-08-23
Updates for 404 / deadlinks

alx-a:
going through 404 links, #204, also commenting there with some ideas on tackling the issue long-term. Details in the 404 fixes often need discussing, so making small individual commits. 

Looks like rigetti  magicl high-level got broken down into smaller files and the original is gone / 404. I'm linking to the dir and doc for now. 

This is ready for merge.

### DIFF
--- a/arrays.md
+++ b/arrays.md
@@ -72,7 +72,7 @@ are given on the [github web pages](https://github.com/rigetticomputing/magicl).
 Low-level routines wrap foreign functions, so have the Fortran names
 e.g `magicl.lapack-cffi::%zgetrf`. Higher-level interfaces to some of
 these functions also exist, see the
-[source code](https://github.com/rigetti/magicl/blob/master/src/high-level/high-level.lisp).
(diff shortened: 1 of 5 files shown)


bendudson:
Thanks! This change looks good to me. It might good to add a link to NumCL to this page (https://github.com/numcl/numcl), and think about which library this page should point new users to. At the moment it's mainly using examples from `array-operations` because that seemed relatively straightforward and covered a lot of common use-cases. Has NumCL now reached a state where it can be recommended as the place new users should start?

alx-a:
Ben, 
1. suggest a NumCL library, do I understand correctly? I won't be able to test for a bit, maybe others have opinions. What's yours?
2.  From my perspective that raises the question of whether by *new users* the text would target individuals experienced with Numpy but not CL, vice versa, or neither.  I'm not sure. If we outgo from Numpy literacy, the recipes could certainly be to the point, with less explanation. Perhaps a good first draft direction?

vindarel:
Thanks again!

PR #414, opened 2021-09-17
testing: how to run tests on the terminal with the right exit code

vindarel:
Frankly, I have struggled way too much with that exit code… I'll learn better practices with the greatest interest, if you have suggestions.


vindarel:
And `run-fiveam` usage is not very clear.

vindarel:
Asked on Discord and they do more or less the same.

PR #419, opened 2021-10-31
strings: Add more format examples

lnguyen46:


strings.md
```diff
@@ -1009,6 +1010,31 @@ With `~2,2f`:
 
 And we're happy with this result.
 
+### Iteration
```
vindarel, line 1013: please add a newline after a title.

strings.md
```diff
@@ -1009,6 +1010,31 @@ With `~2,2f`:
...
+Create a string from a list with iteration construct `~{str~}`:
```
vindarel, line 1014: I am a bit confused about "str" in here. What would be best? It is a format directive. Use some `<directive>` placeholder (not best), `…` or nothing?

strings.md
```diff
@@ -1009,6 +1010,31 @@ With `~2,2f`:
...
+~~~lisp
```
vindarel, line 1015: and also add a newline before the `~~~lisp` code snippet, it is necessary for the epub generation.

Necessary for all the snippets below.

strings.md
```diff
@@ -1009,6 +1010,31 @@ With `~2,2f`:
...
+
+using `~^` to avoid printing the comma and space after the last element:
+~~~lisp
+(format t "~{~A~^, ~}" '(a b c))
+;; A, B, C
+~~~
+
+`~:{str~}` is similar but for a list of sublists:
+~~~lisp
+(format nil "Pairs:~:{ <~S,~S>~} ." '((a 1) (b 2) (c 3)))
```
vindarel, line 1028: I think using real words instead of two more symbols (< and >) would be more readable. As a newcomer we might not understand straight if < is part of the directive or not.

strings.md
```diff
@@ -1022,6 +1048,12 @@ comma-separated prefix parameters:
 ;; "foo                           "
 ~~~
 
+or to create a string having repititive characters by combining with `~:*`:
+~~~lisp
+(format t "~v@{~A~:*~}" 5 "*")
```
vindarel, line 1053: so this is like `(make-string 5 :initial-element #\*)`? Then is it worth showing this obscure format directive? (if you feel like yes, then yes)

strings.md
```diff
@@ -1045,6 +1077,29 @@ Of course, it is always possible to format a format string beforehand:
...
+(format nil "~[dog~;cat~;bird~:;default~]" 9)
+;; "default"
+~~~
+
+Combine it with `~:*` to implement irregular plural:
+~~~lisp
+(format nil "I saw ~r el~:*~[ves~;f~:;ves~]." 0) ==> "I saw zero elves."
+(format nil "I saw ~r el~:*~[ves~;f~:;ves~]." 1) ==> "I saw one elf."
+(format nil "I saw ~r el~:*~[ves~;f~:;ves~]." 2) ==> "I saw two elves."
+~~~
```
vindarel, line 1102: TIL, thanks!

vindarel (changes requested):
Thanks!

strings.md
lnguyen46: I have deleted it.

strings.md
lnguyen46: changed!

strings.md
lnguyen46: I don't know what is the best in this case.

The "str" here is actually copied from documentation in LispWorks HyperSpec.

vindarel:
Thank you!

PR #422, opened 2021-11-14
Complete bodge-ui example

dpflug:
The Nuklear/bodge-ui example seems to be incomplete. This extends it to match the bodge-ui simple example.


vindarel:
Thanks a lot (and sorry for the delay). I'm glad you had the example working. I am hesitant to leave this example here, because the developer doesn't think it's a good idea to advertise the bindings, as they are too low-level and were more an exercise for him than a goal of its own.

PR #431, opened 2022-01-20
Adds guix shell instructions

jgarte:
Hi, 

This adds instructions on how to load the manifest with [guix shell](https://guix.gnu.org/en/blog/2021/from-guix-environment-to-guix-shell/).

This would work on any GNU/Linus distro as long as you have GNU Guix installed.

Debian has GNU Guix officially packaged:

```
sudo apt install guix
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,10 @@ The main goal is making the Cookbook more modern and more accessible in addition
 [contributing]: CONTRIBUTING.md
 [bundler-v2]: https://stackoverflow.com/questions/54087856/cant-find-gem-bundler-0-a-with-executable-bundle-gemgemnotfoundexceptio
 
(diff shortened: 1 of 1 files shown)


vindarel:
Nice, thanks!

vindarel:
(and very simple. Liking guix so far as an alternate package manager, although it takes a lot of disk space)

jgarte:
> (and very simple. Liking guix so far as an alternate package manager, although it takes a lot of disk space)

@vindarel 

Here's a starter CL app with a proposed guix workflow:

https://git.genenetwork.org/GeneNetwork/cl-gn

I'm curious if you have any ideas how we can make the Guix experience better or similar to the quicklisp experience?

PR #440, opened 2022-04-06
remove empty leading newline in code block

augustfengd:
Hello! I was reading through the page and this looked a bit odd.


vindarel:
indeed, thanks!

PR #443, opened 2022-04-19
Add example on how to join list of strings

MichaelAnckaert:
This PR includes an example on how to join a list of strings to a string, delimited with a character. 

```lisp
* (join '("foo" "bar" "baz") #\-)
"foo-bar-baz"
```

### DIFF
--- a/strings.md
+++ b/strings.md
@@ -829,6 +829,31 @@ represented as quotients of two integers even when converted to strings.
 *
 ~~~
 
+## Converting a List to a String
+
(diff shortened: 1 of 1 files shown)


pnathan:
consider using FORMAT as per either mine or R. Joswig's answer here: https://stackoverflow.com/questions/8830888/whats-the-canonical-way-to-join-strings-in-a-list

vindarel:
Hi, thanks for sending a PR. I wish we could use shorter examples, especially when we have `str:join` or format.

PR #446, opened 2022-04-24
Update and add links for vim editor support

Jach:
Though there haven't been many changes to Slimv since the 2017 release on vim.org, the biggest change being some basic stepper support, I think it's more expected these days to have links to a project's development home.

I also think other vim users will be interested in the linked blog post to help them decide which of the two plugins to try.

### DIFF
--- a/editor-support.md
+++ b/editor-support.md
@@ -50,7 +50,7 @@ See ["Using Emacs as an IDE"](emacs-ide.html).
 
 ## Vim & Neovim
 
-[Slimv](https://www.vim.org/scripts/script.php?script_id=2531) is a full-blown
+[Slimv](https://github.com/kovisoft/slimv) is a full-blown
 environment for Common Lisp inside of Vim.
 
 [Vlime](https://github.com/vlime/vlime) is a Common Lisp dev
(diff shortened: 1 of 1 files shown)


vindarel:
Good points. We avoid linking to blog posts in the middle of the page though. We either write content, either use the section for that at the bottom.

Jach:
Yeah, agreed it's better to have self-contained content. But some already existing pages are just really useful and not directly findable from existing links (like it'd be redundant to link to slimv's own tutorial since it's part of its readme), and I noticed the link just above in the emacs section that goes out to wikemacs indicating an exception to self-contained content.

I can remove the blog link completely if you think that's best and just make this PR about updating the slimv link, but I'm happy/somewhat expected to bikeshed a bit on the proper formatting to include the blog post. It wasn't clear to me that the screenshot really denoted a 'sectioning' so it made most sense initially to put the link close to the two things it's about. To check that I understand what you mean by the section at the bottom, do you just mean adding something like:

> See also:
> * link

Below the slimv\_box line and above the Atom section? Or make an explicit subsection there like "Using Slimv and Vlime with Vim" and putting the link inside that, and ideally one day replacing it with or adding a self-contained cookbook page?

vindarel:
yes, can you add this "See also… link" at the end of the Vim entry, above he Atom section?

jgarte:
@Jach @vindarel 

Hi, is this still being worked on?

Jach:
I made the changes requested, so nothing else from my side I think

vindarel:
Thanks both!

PR #448, opened 2022-05-10
Added 'setting the current working directory' to 'Files and Directories'

chuchana:
with `uiop/os:chdir path`


vindarel:
Hey, the function is exported and available with `uiop:chdir` too, we should mention that.

chuchana:
Right, I'll change that. `uiop:chdir` didn't work when I tried it the other day, but I guess it was just a problem with my system.

PR #450, opened 2022-06-10
Update default.html: add favicon

tinkerttoy:
I've repurposed the logo from the sidebar as a favicon. This should make it much easier to pick out the Cookbook from among tabs and bookmarks


vindarel:
yay, thank you!

Issue #425, opened 2021-11-27
Pathname recipes


contrapunctus-1:
I was thinking of adding recipes for some common pathname operations, but the files page says -

> In this chapter, we use mainly [namestrings](http://www.lispworks.com/documentation/HyperSpec/Body/19_aa.htm) to [specify filenames](http://www.lispworks.com/documentation/HyperSpec/Body/19_.htm). The issue of [pathnames](http://www.lispworks.com/documentation/HyperSpec/Body/19_ab.htm) needs to be covered separately.

...so I guess I'll make a new page (and thus announce it here). Although, to be honest, "Files and Directories" is actually where I expect to have path operations too...

vindarel:
yeah, a new section on that page makes sense. Any recipe on pathnames is very welcome!

PR #454, opened 2022-06-29
Create pathname components section

contrapunctus-1:
An attempt at tackling #425 . The "parent directory" recipe is not perfect - it does not work for files.


vindarel:
thanks!

PR #458, opened 2022-08-21
First Draft: Add FFI subsection on math.h

### LINKED ISSUE #33: Use CFFI library in "FFI" Section
UFFI is an abstraction layer over all well-known Common
Lisp FFI implementations. It may be interessting to add
an example using this layer in the FFI section of the
CL Cookbook.

http://uffi.med-info.com/

[Frederic Brunel @ http://sourceforge.net/p/cl-cookbook/feature-requests/2/]

jgarte:
Closes #33

### DIFF
--- a/ffi.md
+++ b/ffi.md
@@ -4,6 +4,97 @@ title: Foreign Function Interfaces
 
 The ANSI Common Lisp standard doesn't mention this topic. So almost everything that can be said here depends on your OS and your implementation.
 
+### Example: Calling mathematical functions from C's `math.h` header file.
+ 
+Let's use `defcfun` to interface with the foreign [ceil](https://en.cppreference.com/w/c/numeric/math/ceil) C function from `math.h`.
+
+[defcfun](https://cffi.common-lisp.dev/manual/html_node/defcfun.html) is a macro in the cffi library that generates a function with the name you give it in your local scope.
+
+CL-USER> CL-USER> (cffi:defcfun ("ceil" c-ceil) :double (number :double))
+
+Here is the above function macroexpanded with `macrostep-expand`:
+
+~~~lisp
+(progn
+	  nil
+	  (defun c-ceil (number)
+	    (let ((#:g312 number))
+	      (cffi-sys:%foreign-funcall "ceil" (:double #:g312 :double) :convention
+					 :cdecl :library :default))))
+~~~
+
+As we can see from the above macroexpansion, and ignoring other details, a function is created with the builtin `defun` that is called `c-ceil` and that takes exactly one argument: `number`.
+
+The reason we called it `c-ceil` and not `ceil` is because sbcl already implements a `ceil` function and our generated FFI function produced from the `defcfun` macro would clash with it.
+
+Now that we have a c-ceil function from `math.h`, let's use it!
+
+~~~lisp
+CL-USER> (c-ceil 5.4d0)
+6.0d0
+~~~
+
+As you can see, it works! The double gets rounded up to `6.0d0` as expected.
+
+Let's try another one! This time, we'll use [floor](https://en.cppreference.com/w/c/numeric/math/floor) from `math.h`:
+
+~~~lisp
+CL-USER> (cffi:defcfun ("floor" c-floor) :double (number :double))
+C-FLOOR
+CL-USER> (c-floor 5d0)
+5.0d0
+CL-USER> (c-floor 5.4d0)
+5.0d0
+~~~
+
+Great!
+
+One more, let's try sqrt from math.h:
+
+~~~lisp
+CL-USER> (cffi:defcfun ("sqrt" c-sqrt) :double (number :double))
+C-SQRT
+CL-USER> (c-sqrt 36)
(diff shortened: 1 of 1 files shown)


jgarte:
note to self:

[char](https://github.com/charje/) says to

> (coerce 12.34 'double-float)
> I would include that in the c-* functions

ffi.md
```diff
@@ -4,6 +4,109 @@ title: Foreign Function Interfaces
 
 The ANSI Common Lisp standard doesn't mention this topic. So almost everything that can be said here depends on your OS and your implementation.
 
+### Example: Calling mathematical functions from C's `math.h` header file.
+ 
+This tutorial will use Guix to install cffi in an isolated shell environment. The REPL session below was tested in a sly repl.
```
vindarel, line 9: As a reader, I wonder if guix is necessary, why, and if I wanted to do without what would it take or what would be the limitations (in a few words).

vindarel:
Nice. First draft, so not ready to merge?

ffi.md
jgarte: Guix is definitely not necessary here. I can remove that and send an update. I just made a quick draft of this.

I will say, that using Guix introduces a different workflow than using quicklisp, for example, because libraries can not be dynamically installed, updated, reloaded in a REPL without restarting emacs (there's a workaround).

jgarte:
@vindarel I pushed an update to remove the Guix specific instructions.

ffi.md
```diff
@@ -4,6 +4,97 @@ title: Foreign Function Interfaces
 
 The ANSI Common Lisp standard doesn't mention this topic. So almost everything that can be said here depends on your OS and your implementation.
 
+### Example: Calling mathematical functions from C's `math.h` header file.
+ 
+Let's use `defcfun` to interface with the foreign [ceil](https://en.cppreference.com/w/c/numeric/math/ceil) C function from `math.h`.
+
+[defcfun](https://cffi.common-lisp.dev/manual/html_node/defcfun.html) is a macro in the cffi library that generates a function with the name you give it in your local scope.
+
+CL-USER> CL-USER> (cffi:defcfun ("ceil" c-ceil) :double (number :double))
```
Bike, line 13: Duplicate CL-USER>

ffi.md
```diff
@@ -4,6 +4,97 @@ title: Foreign Function Interfaces
...
+
+Here is the above function macroexpanded with `macrostep-expand`:
+
+~~~lisp
+(progn
+	  nil
```
Bike, line 19: Indentation is messed up

ffi.md
```diff
@@ -4,6 +4,97 @@ title: Foreign Function Interfaces
...
+	  (defun c-ceil (number)
+	    (let ((#:g312 number))
+	      (cffi-sys:%foreign-funcall "ceil" (:double #:g312 :double) :convention
+					 :cdecl :library :default))))
+~~~
+
+As we can see from the above macroexpansion, and ignoring other details, a function is created with the builtin `defun` that is called `c-ceil` and that takes exactly one argument: `number`.
```
Bike, line 26: We don't usually refer to "builtins" in Lisp, and really I don't think you need to explain what `defun` is. I would say "a function `c-ceil` is defined that takes exactly one argument"

ffi.md
```diff
@@ -4,6 +4,97 @@ title: Foreign Function Interfaces
...
+
+The reason we called it `c-ceil` and not `ceil` is because sbcl already implements a `ceil` function and our generated FFI function produced from the `defcfun` macro would clash with it.
```
Bike, line 28: Er, does SBCL actually have a `ceil`? There's the standard `cl:ceiling`, but in 2.2.9 I don't see any `ceil` defined, let alone in a cl-user-accessible package.

ffi.md
```diff
@@ -4,6 +4,97 @@ title: Foreign Function Interfaces
 
 The ANSI Common Lisp standard doesn't mention this topic. So almost everything that can be said here depends on your OS and your implementation.
 
+### Example: Calling mathematical functions from C's `math.h` header file.
+ 
+Let's use `defcfun` to interface with the foreign [ceil](https://en.cppreference.com/w/c/numeric/math/ceil) C function from `math.h`.
+
+[defcfun](https://cffi.common-lisp.dev/manual/html_node/defcfun.html) is a macro in the cffi library that generates a function with the name you give it in your local scope.
```
Bike, line 11: `defcfun` defines a global function, not one in a local scope.

Bike (commented):
Some nitpicks, since you asked

vindarel:
Suggestions: show the function signature of C `ceil`, what it accepts as input and what it outputs, so when you show 

    (cffi:defcfun ("ceil" c-ceil) :double (number :double))

you can also explain that this is the input, this is the output and what is the overall syntax of `defcfun`.

jgarte:
@vindarel  @Bike 

Thanks for the suggestions! They are much appreciated. Give me a few days (maybe sooner) and I'll push some more updates. I'm just busy at the moment finishing out work before Thanksgiving break :)

vindarel:
hey @jgarte it would be really great to finish this.

PR #462, opened 2022-09-30
Add "easy-routes" package

blackeuler:
This adds the `easy-routes` package to the initial installation list for the libraries needed.


vindarel:
Thanks!

PR #464, opened 2022-11-09
Update clos.md

createyourpersonalaccount:
Add an example on how to get the documentation of a slot.

### DIFF
--- a/clos.md
+++ b/clos.md
@@ -329,7 +329,13 @@ class (whether or not those instances exist yet).
 
 #### Slot documentation
 
-Each slot accepts one `:documentation` option.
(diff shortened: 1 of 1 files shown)


clos.md
```diff
@@ -329,7 +329,13 @@ class (whether or not those instances exist yet).
 
 #### Slot documentation
 
-Each slot accepts one `:documentation` option.
+Each slot accepts one `:documentation` option. To obtain its documentation via `documentation`, you need to obtain the slot object. This can be done compatibly using an implementation such as [closer-mop](https://github.com/pcostanza/closer-mop). For instance:
```
vindarel, line 332: > using an implementation

is "implementation" right? I use implementation for SBCL, CCL… I'd use "compatibility library" or "library". Thanks!

vindarel (changes requested):

clos.md
createyourpersonalaccount: You're right! Sometimes they speak of MOP implementations, but that's not what closer-mop is. I mixed up the wording.

PR #472, opened 2022-11-29
debugging: trace options, more on stepping

vindarel:
https://vindarel.github.io/cl-cookbook/debugging.html


vindarel:
> I'm making a short screencast

Here it is and it is über cool^^ [Debugging Lisp: fix and resume a program from any point in the stack](https://www.youtube.com/watch?v=jBBS4FeY7XM) - with Emacs and Lem.

PR #475, opened 2022-12-27
packages: example with add-package-local-nickname

jgarte:
@vindarel 

Hi, 

this adds an example with `add-package-local-nickname` which I find to be very useful for trying out `PLN`s.

WDYT

### DIFF
--- a/packages.md
+++ b/packages.md
@@ -143,6 +143,16 @@ provide nice nicknames.
 
 Many implementations (SBCL, CCL, ECL, Clasp, ABCL, ACL, LispWorks >= 7.2…) support Package Local Nicknames (PLN).
 
+To use a `PLN` you can simply do the following, for example, if you'd like to try out a `PLN` in an ad-hoc fashion:
+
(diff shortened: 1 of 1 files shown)


packages.md
```diff
@@ -143,6 +143,16 @@ provide nice nicknames.
 
 Many implementations (SBCL, CCL, ECL, Clasp, ABCL, ACL, LispWorks >= 7.2…) support Package Local Nicknames (PLN).
 
+To use a `PLN` you can simply do the following, for example, if you'd like to try out a `PLN` in an ad-hoc fashion:
+
+~~~lisp
+(add-package-local-nickname :a :alexandria)
```
vindarel, line 149: it's in uiop

vindarel (changes requested):

vindarel:
Awesome! TIL.

packages.md
jgarte: I'll add it today and tag you when I push. Thanks!

PR #485, opened 2023-02-19
Fix instructions for local CLHS  lookup in Emacs

ralfdoering:
The instructions missed the call to " (clhs:install-clhs-use-local)" and referenced the wrong directory for the .el file.


kilianmh:
Thank you for your contribution @ralfdoering!

kilianmh:
Actually, this was my first PR merging ;) 
I thought it would be committed by default with only one commit. From now on it will be that way.

vindarel:
First PR merging on GitHub? Cool, welcome :)

PR #489, opened 2023-03-20
[HOLD] packages: add instructions on using packages

ChanderG:
Added a few instructions on getting started with using packages. This is information I needed to just write small scripts - getting started with CL and could find only with some trial/error.

### DIFF
--- a/packages.md
+++ b/packages.md
@@ -4,6 +4,44 @@ title: Packages
 
 See: [The Complete Idiot's Guide to Common Lisp Packages][guide]
 
+## Using packages
+
+To install and use a package with a single command, use `ql:quickload` as follows:
+~~~lisp
+(ql:quickload :package)
+
+(package:function ...)
+~~~
+where package can be specified as a string or as a keyword. Packages available using Quicklisp can be found on [https://quickdocs.org/](Quickdocs).
+
+Instead, if you have installed the package separately, you can use:
+~~~lisp
+(require :package)
+
+(package:function ...)
+~~~
+As before you can use the keyword `:package` or string `"package"`. Note that if the package is not found, it will fail to compile.
+
+You can import individual symbols from a package:
+~~~lisp
+(require :package)
+(import 'package:function)
+
+; can now use it directly
+(function ...)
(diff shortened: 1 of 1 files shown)


packages.md
```diff
@@ -4,6 +4,44 @@ title: Packages
 
 See: [The Complete Idiot's Guide to Common Lisp Packages][guide]
 
+## Using packages
+
+To install and use a package with a single command, use `ql:quickload` as follows:
+~~~lisp
+(ql:quickload :package)
```
vindarel, line 11: I think this sentence misleads packages with systems. 

Also quickload is shown on systems.md / loading a system.

and how to access symbols with `package:function` is shown below at "accessing symbols from a package".

The current recipe shows a package definition, that effectively creates a package.  I wouldn't start this section with a quickload example.

Could you rephrase, or add an introduction explaining the difference between systems and packages, linking to the systems page if necessary?

packages.md
```diff
@@ -4,6 +4,44 @@ title: Packages
 
 See: [The Complete Idiot's Guide to Common Lisp Packages][guide]
 
+## Using packages
+
+To install and use a package with a single command, use `ql:quickload` as follows:
+~~~lisp
```
vindarel, line 10: please add a blank line before all the `~~~lisp`, thanks.

vindarel (changes requested):

ChanderG:
> Hi @ChanderG, maybe you needed a quick way to see how to use external libraries ?

Yes - that's exactly right. Maybe this section should go somewhere else and not in this page on packages?

packages.md
ChanderG: Yes - to be honest - I haven't yet grasped the difference between those terms.

vindarel:
I sent changes that should answer your concerns. I added sub-bullets in the index to see "install libraries" more quickly and I augmented the content with your input from this PR (mainly the packages page).

ChanderG:
@vindarel Your changes seem to cover everything I was looking for. Thanks!
Will close this PR now.

PR #492, opened 2023-04-11
index: decompose in four topics

vindarel:
Add 4 sections: getting started, language basics, advanced topics, outside world.


jgarte:
LGTM! Looks way more tidy. Thanks for doing that!

PR #494, opened 2023-04-14
gui.md: fix typo in nodgui demo function call

grobe0ba:


vindarel:
Oh, really? It would be nice to export this symbol. Thanks.

PR #495, opened 2023-04-24
Updated sbcl core compression session to zstd

funk443:
I am not sure what you guys used to test the difference between image sizes, so I leave it unmodified.


vindarel:
Thanks. The size difference was nothing scientific but an observation with one of my apps, with a few dependencies.

vindarel:
yes, and both? Because I like to see that a CL binary, even with dozens of dependencies, is in the ballpark of 35-40MB. That's interesting to know for outsiders, who are afraid (with reasons or not) by the size of a "hello world" binary.

PR #496, opened 2023-04-25
Add compression level and size comparison

funk443:
Two nice little tables to demonstrate how compression levels and file sizes are related

### DIFF
--- a/scripting.md
+++ b/scripting.md
@@ -332,7 +332,27 @@ In SBCL, we would give an argument to `save-lisp-and-die`, where
 
 > may be an integer from -7 to 22, corresponding to zstd compression levels, or t (which is equivalent to the default compression level, 9).
 
-We experienced a 1MB difference between levels -1 and 9.
+For a simple "Hello, world" program:
+
+```
+| Program size | Compression level   |
+|--------------|---------------------|
+| 46MB         | Without compression |
+| 22MB         | -7                  |
+| 12MB         | 9                   |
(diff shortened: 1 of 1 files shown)


vindarel:
Excellent, thanks!

re presentation: it's "MB" in english (yes?), not "MiB".

I read better the columns in the order "compression level" -> "program size"

The markdown table doesn't render well in the epub and pdf :( Could use transform to a HTML table? Like so:

```html

<style>
table, th, td {
  border: 1px solid black;
  border-collapse: collapse;
}
th, td {
  padding: 5px;
}
th {
  text-align: left;
}
</style>

<table>
  <tbody>
    <tr>
      <td>Compression level</td>
      <td>Program size</td>
    </tr>
    <tr>
      <td>Without compression</td>
      <td>46MB</td>
    </tr>
  </tbody>
</table>
```

thanks again

funk443:
Hello!

"MiB" stands for "MebiByte", while "MB" stands for "MegaByte".  I use "MiB" because that's how my system measures the file size.  

And I did put the table in a code block, so that might not be a problem?

vindarel:
TIl about this notation. I used them interchangeably.

sure a simpler code block will do the job very well.

funk443:
So should I convert them into MB?

vindarel:
"should", no, but maybe yes, as it is more common? I thought MiB was a typo, but since it's correct I was accepting it.

PR #502, opened 2023-07-22
index.md: Change "Threads" page name to "Concurrency and Parallelism" 

### LINKED ISSUE #501: Change "Threads" page name to "Concurrency and parallelism"
At least for me, the name "Threads" makes it seem like the page only talks about thread-based, preemptive multitasking recipes.

But in reality, the page goes over the [lparallel](https://lparallel.org/) library, which supports many other models of multitasking. So I think a name such as "Concurrency and parallelism" is more appropriate for the page.

able-j:
closes #501


vindarel:
Thanks :)

PR #508, opened 2023-10-30
Unit test example does not work (testing.md)

### LINKED ISSUE #490: Unit test example does not work
When trying to define tests as shown in `testing.md`, the simple math example does not work; you just get:

```
The variable SIMPLE-MATHS is unbound.
```

able-j:
Closes #490


vindarel:
Thanks!

PR #514, opened 2023-12-04
Change unfortunate char code in string example

dangerdyke:
The number 1488, used in the strings page as an example character code, is often used as a covert symbol by nazi groups ([source](https://www.dictionary.com/e/politics/1488/)). I thought it best to change it to something innocuous, since it's a bad look regardless of the original intention behind the example.

### DIFF
--- a/strings.md
+++ b/strings.md
@@ -169,11 +169,11 @@ value 255). Notice the Lisp reader can round-trip characters by name.
 * (char-code #\LATIN_CAPITAL_LETTER_E_WITH_GRAVE)
 
 200
-* (code-char 1488)
-#\HEBREW_LETTER_ALEF
+* (code-char 2048)
+#\SAMARITAN_LETTER_ALAF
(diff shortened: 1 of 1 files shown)


pnathan (approved):

pnathan:
Awkward and ironic, given the character point used.

vindarel:
f*. thank you

`git blame` someone?

pnathan:
https://www.compart.com/en/unicode/U+05D0 

looks like somehow the Alef / Aleph range starts at the unfortunate number. Strange business...

dangerdyke:
thanks for the quick merge :)

> `git blame` someone?

I took a look through the blame when I first noticed it, looks like that example was added in 2007, before the cookbook was on GH: https://github.com/LispCookbook/cl-cookbook/commit/39914b61ea9fbe1a617e2262794fa66b31fb451f

PR #519, opened 2024-01-11
Show how to read arbitrary bytes into a string for SBCL.

Kevinpgalligan:
There's no entry for SBCL on this page, I got stuck on this for a while.


vindarel:
Thank you

PR #526, opened 2024-01-23
Describe how to achieve Python-style lists.

Kevinpgalligan:
The data structures page doesn't describe in detail how to make arrays that grow their capacity as more elements are added, like Python lists. So I've added such a description.

Maybe it's too Python-specific --  I'm open to suggestions of how to present it better. If there's a better way to pop from a vector while erasing the element, or if there's a way to ensure that an error is raised when trying to access an already-popped element through `(aref *v* index-of-popped-element)`, then I'm also open to suggestions on that front.

### DIFF
--- a/data-structures.md
+++ b/data-structures.md
@@ -903,6 +903,26 @@ points to.
 
 and see also the *sequence* functions.
 
+### "I just want Python-style lists!!!"
+The following creates an empty array that can be pushed to and popped from arbitrarily, growing its storage capacity as needed, as is the case for Python lists.
+
+~~~lisp
+(defparameter *v* (make-array 0 :fill-pointer t :adjustable t))
+(vector-push-extend 42 *v*)
+(vector-pop *v*)
+~~~
+
+Note that `vector-pop` does not erase the element in the underlying storage array, as can be confirmed by calling `(array-storage-vector *v*)` -- in this example, it returns `#(42)`. So to avoid leaving garbage around, the following wrapper can be defined instead. 
+
+~~~lisp
+(defun vector-erasepop (v)
+  (let ((x (vector-pop v)))
+    (setf (aref v (fill-pointer v)) nil) ; or 0
+    x))
+~~~
+
+A final warning: `(aref *v* 0)` will still return `nil` rather than raising an error.
+
 ### Transforming a vector to a list.
 
 If you're mapping over it, see the `map` function whose first parameter
--- a/data-structures.md
+++ b/data-structures.md
@@ -890,38 +890,34 @@ vector._
 ;; => #(1 2 3)
 ~~~
 
+The following interface is available for vectors (or vector-like arrays):
 
-`vector-push` *(foo vector)*: replace the vector element pointed to by
-the fill pointer by foo. Can be destructive.
-
-`vector-push-extend` *(foo vector [extension-num])*t
-
-`vector-pop` *(vector)*: return the element of vector its fill pointer
-points to.
-
-`fill-pointer` *(vector)*. `setf`able.
+* `vector-push` *(new-element vector)*: replace the vector element pointed to by the fill pointer by `new-element`, then increment the fill pointer by one. Returns the index at which the new element was placed, or NIL if there's not enough space.
+* `vector-push-extend` *(new-element vector [extension-num])*: like `vector-push`, but if the fill pointer gets too large then the array is extended using `adjust-array`.
+* `vector-pop` *(vector)*: decrement the fill pointer, and return the element that it now points to.
+* `fill-pointer` *(vector)*. `setf`able.
 
 and see also the *sequence* functions.
 
-### "I just want Python-style lists!!!"
-The following creates an empty array that can be pushed to and popped from arbitrarily, growing its storage capacity as needed, as is the case for Python lists.
+The following shows how to create an array that can be pushed to and popped from arbitrarily, growing its storage capacity as needed. This is roughly equivalent to a `list` in Python, an `ArrayList` in Java, or a `vector<T>` in C++ -- though note that elements are not erased when they're popped.
 
 ~~~lisp
-(defparameter *v* (make-array 0 :fill-pointer t :adjustable t))
-(vector-push-extend 42 *v*)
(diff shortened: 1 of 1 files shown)


vindarel:
yes the wording at least is a bit too Python-specific.

Then what's the issue? I think it isn't well exposed. CL lists are linked lists and "grow", the wording implies they don't. Maybe your issue is the access time in O(n)? 

> garbage

this is garbage for pythonic eyes but expected CL behavior right?

(edit)

> `vector-pop` does not erase the element in the underlying storage array

ok got it. You can explain that for vector-pop first.

>  rather than raising an error

why should it return an error, still for pythonic eyes?

---

So is it the combination of `:fill-pointer t :adjustable t` that you find not described enough?

lassik:
Hello.

What problem is this trying to solve? (What is the application doing? Is there an application?)

Is the motivation as simple as, "`push` grows the list in reverse and that looks weird"?

"I just want Python-style lists" does not seem like a productive goal, though understanding extendable vectors certainly is one.

Perhaps the most immediately useful application of extendable vectors is that you can append to strings destructively. Python strings can grow by default, whereas CL strings cannot.

Kevinpgalligan:
> So is it the combination of :fill-pointer t :adjustable t that you find not described enough?

Pretty much.

I was searching (and asking in IRC) "how to do Python-style lists in CL" (or Java-style ArrayLists, or C++-style vectors) and it took me a long time to figure it out. If I want an array that grows arbitrarily then it's not exactly clear how to do that from the Cookbook page. Lists (i.e. cons cells) might not meet someone's needs because of O(n) insertion and indexing. Their interface is also awkward when you're coming from Python's:

```
lst = []
lst.append(1)
lst.pop()
```

(Or equivalents in other imperative languages). I know we have `push` and `pop`, but they operate on "places" so they don't work in all contexts; also, it's annoying having to `reverse` lists all the time. So my first intention was to explain "here's how you get this data structure that you probably expect from other languages".

Secondly, if someone is used to Python (and frankly any other programming language), they would expect the following code to erase the reference to `42` in `v` and raise a "list index out of range" error when the user attempts to `aref` it:

    (let ((v (make-array 0 :fill-pointer t :adjustable t)))
      (push-vector 42 v)
      (pop-vector v)
      (aref v 0))

...but instead, you get back `42`. I wanted to cover these gotchas in the Cookbook.

I can remove the Python-specific language, though, and just move some of the text / examples I wrote under the existing `Vectors` heading -- sound good?

lassik:
The lingering elements after the fill pointer are normally not displayed:

```Common Lisp
> (defvar v (make-array 10 :fill-pointer 0))
V
> (dotimes (i 5 v) (vector-push i v))
#(0 1 2 3 4)
> (dotimes (_ 3 v) (vector-pop v))
#(0 1)
```

Kevinpgalligan:
That's why I find it especially pernicious: let's say someone is `vector-pop`-ing massive objects from their array. They might expect those objects to be garbage-collected, but in fact there's a hidden reference keeping them alive.

lassik:
> Lists (i.e. cons cells) might not meet someone's needs because of O(n) insertion and indexing.

So they need a deep stack with frequent random access?

(Linked lists are not slow for few elements or infrequent access.)

> Their [cons cells] interface is also awkward

Agreed.

> it's annoying having to `reverse` lists all the time.

Agreed.

Sometimes `collect` helps: `(loop for x from 1 to 10 collect x)`

> They might expect those objects to be garbage-collected, when in fact there's a hidden reference keeping them alive, leading to out-of-memory errors.

True.

Mixing `aref` and `vector-pop` is probably not a good idea.

As you say, fill pointers are confusing. People generally avoid them and use linked lists for stacks.

If you're implementing a more complex data structure (e.g. queue, tree, heap) from a textbook, those are generally wrapped in a structure or a class.

lassik:
The higher-level concern here is that vectors are not lists in CL. If your code interfaces with other people's code, they will expect lists to be linked lists.

The only general solution here is:

* Accept that CL lists and vectors are as they are, warts and all.

* Prefer things like `map` and `mapcar` and `reduce` and `loop collect` to avoid having to `cons` and `reverse`.

* Your public functions should take linked lists and return linked lists unless exceptional reason to do otherwise.

* In private code (e.g. the implementation of an elaborate data stucture), vectors are fine.

vindarel:
> Your public functions should take linked lists and return linked lists unless exceptional reason to do otherwise.

mmh, I wouldn't make a difference between public/private code. It's better to handle a non-small number of objects with arrays rather than lists.

> I can remove the Python-specific language, though, and just move some of the text / examples I wrote under the existing Vectors heading -- sound good?

@Kevinpgalligan sounds good. You could compare with how python works, but as a side note, not so central. Do you find something in Alexandria, Serapeum or in another library that would do what you seek?

https://github.com/ruricolist/serapeum/blob/master/REFERENCE.md#vectors

Kevinpgalligan:
I gave it another pass, let me know if that works or if you have any further thoughts :)

data-structures.md
```diff
@@ -890,18 +890,34 @@ vector._
 ;; => #(1 2 3)
 ~~~
 
+The following interface is available for vectors (or vector-like arrays):
 
-`vector-push` *(foo vector)*: replace the vector element pointed to by
-the fill pointer by foo. Can be destructive.
+* `vector-push` *(new-element vector)*: replace the vector element pointed to by the fill pointer by `new-element`, then increment the fill pointer by one. Returns the index at which the new element was placed, or NIL if there's not enough space.
+* `vector-push-extend` *(new-element vector [extension-num])*: like `vector-push`, but if the fill pointer gets too large then the array is extended using `adjust-array`.
```
vindarel, line 896: explain "extension-num" ?

vindarel:
It's perfect! and really helpful. I invite you to stay around and keep an eye open for more possible enhancements, specially in the vector/math/numbers area ; )

Kevinpgalligan:
Sure! I think the Cookbook is a really great resource, and I'm trying to develop the good habit of sharing my learnings whenever I figure something out!

PR #533, opened 2024-02-15
Added Dockerfile for running Jekyll

em7:
Hi,

I had troubles installing older Ruby onto a newer Linux-based OS (Ubuntu 22.04 Jammy). I created a Docker container with Jekyll. I hope you can find it useful, too.

(Gem didn't list `bundler` version 2.1.4 as available for installation, therefore there is 2.1.2 in Dockerfile. It spawns a warning but wors just fine.)

Martin

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM ruby:2.5
+
+COPY . /cl-cookbook
+WORKDIR /cl-cookbook
+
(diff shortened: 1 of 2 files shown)


vindarel:
Thank you very much or as Common Lisp would say "THANK YOU VERY MUCH" :p 

One day we'll update the web stack but for now we can concentrate a bit longer on the content without worrying…

Issue #531, opened 2024-02-12
sbcl-librarian: easy build of shared libraries callable from C and Python (SBCL-only)


em7:
Hello,

Since SBCL can be compiled as a shared library (I have tested it on Linux and Windows; unfortunately, I do not have access to a Mac), the documentation on their website seems somewhat limited. It might not be obvious how to put this feature to good use.

SBCL-librarian https://github.com/quil-lang/sbcl-librarian offers an opinionated and semi-automatic way to create a library callable from C and Python.

Do you think that mentioning this project in CL-Cookbook is a good idea? If so, could I try to write a chapter/page?

Thanks!

vindarel:
Absolutely!

(don't miss their example: https://github.com/quil-lang/sbcl-librarian/tree/main/example)

PR #536, opened 2024-03-04
#531 sbcl-librarian: easy build of shared libraries callable from C and Python (SBCL-only)

em7:
Added chapter about SBCL-Librarian


dynamic-libraries.md
```diff
@@ -0,0 +1,312 @@
...
+  char *init_args[] = {"", "--core", core, "--noinform", };
+  if (initialized) return 1;
+  if (initialize_lisp(4, init_args) != 0) return -1;
+  initialized = 1;
+  return 0; }
+~~~
+
+At the top, you'll find several SBCL-related functions, such as `lisp_gc`, which signals to the Lisp garbage collector that it is a good time to run. Then there is a pointer to the `callback_call_callback` function. Finally, the `init` function, which should be run before executing any Lisp code.
+
+Currently, there is no way to de-initialize the Lisp core.
```
vindarel, line 202: Add a timestamp or mention somewhere (github issue) to track a potential update on this? (if it's a goal and being worked on)

em7: Hi, I'm not sure whether they plan to add this functionality. I'm not even sure whether to include this sentence or not, honestly ...

em7:
Hi, I hope this topic might be useful for someone. If other people could proof read it, it would be nice - there's IMO no need to hurry with releasing a chapter on such a niche topic.

I can definitely make a simpler example if it helps or even include more of the example files (or their whole contents). The actual call is in [example.py](https://github.com/quil-lang/sbcl-librarian/blob/main/examples/callback/example.py), the content of which I didn't include in the chapter since it's quite a straightforward Python CFFI call (Python has a module called CFFI which is different from Common Lisp's CFFI :) ).

As for usage, I don't know of any off the top of my head. I used it as a part of my dissertation - we've created a prediction model for high-frequency time series but it needs to be integrated into Python stuff since that's the language other researchers speak (Python is practically a lingua franca of nowadays AI and ML). SBCL-Librarian is also hosted at [Rigetti](https://en.wikipedia.org/wiki/Rigetti_Computing)'s GitHub space, so I suppose they created it for themselves?

vindarel:
> usage

indeed I was thinking about its creation and use in and by quantum companies (also HRL laboratories, where stylewarning now works after Rigetti, if I'm not mistaken (latest Twitter or reddit news)). It's cool to see "real world" usage outside of the creators' hands too.

dynamic-libraries.md
```diff
@@ -0,0 +1,431 @@
...
+
+you can write:
+
+~~~C
+if (ERR_FAIL == cl_function()) {
+~~~
+
+which is more readable.
+
+`define-api` outlines the structure of the library code to be created, specifying the error map, types, functions, and their order (`:literal` is used for comments in this case). The function `call-callback` uses previously defined types for its arguments: the `callback` type for the first argument named `fn` and the `:char-buffer` type for its second argument `out_buffer`. The `:function-prefix` option means the actual name of the exported function will be `callback_call_callback`.
```
vindarel, line 156: "call-callback" and "char-buffer" are used later and shouldn't appear in this sentence at this point?

vindarel:
Wow, great, it's very well laid out.

dynamic-libraries.md
em7: Sorry about that, "copy-pasta" :) I moved the article to an earlier subchapter and forgot to remove the sentence.

em7:
Thanks! I hope this can be a good contribution.

vindarel:
It is an excellent contribution that not a lot of people could write, and that should attract the attention of advanced users and increase the rhythm of CL taking over the world,

thanks for writing to the Cookbook, it really is a nice resource now thanks to everybody.

If you need or want to update or tweak this page once in a while, I can give you push rights so you can do it at your rhythm. (for bigger contributions a PR is still welcome of course)

em7:
Thank you for the Cookbook (and the Udemy course, it's from you as well, right?), both have been invaluable resources!

I'd prefer not to have push rights. Years in the IT industry taught me that even small changes can benefit from having another pair of eyes on them.

vindarel:
Great to hear, thanks for the feedback, yes the Udemy course is from me.

OK!

In case you don't follow reddit, a question and a feedback by stylewarning: https://www.reddit.com/r/Common_Lisp/comments/1c7tfs6/cookbook_building_dynamic_libraries_with/

em7:
Thanks for the link and the heads-up!

I follow the community, I replied to both David and `stylewarning`, unfortunately, I have never tried to compile SBCL on Android. I'll be starting a commercial side-project later this year which will target both desktops and mobile devices. My original idea was to create everything server-side (either with HTMX or Clog for UI), however the ability to have the most of the logic in Common Lisp with native UI is very tempting.

Issue #540, opened 2024-04-26
Fix typos, grammar, etc


cgay:
Hi, is it okay if I fix typos and similar minutiae as I read through the cookbook, and file the resulting PRs against this bug?

vindarel:
Ab-so-lu-te-ly, thank you in advance. I might ping you in future PRs too ; )

cgay:
Cool. Not saying I'll go through the whole thing at once or anything though. :)

BTW / FYI, I was unable to get jekyll running on macOS 14 (Sonoma) due to Ruby gem dependency problems. (I followed the jekyll instructions to install Ruby because the system's Ruby isn't trusted for some reason.) Jekyll doesn't claim to support Sonoma yet....   So I can't test my changes very well...but they'll be simple changes.

vindarel:
Ok, no pb for Jekyll we have a slightly outdated stack. You could also enable Github pages of your fork and check the result there, but as you said there's probably no need to check simple changes.

Oh, did you set up OpenDylan's website? The docs are pleasant to read and the theme is stellar (Sphinx, I see).

cgay:
> Ok, no pb for Jekyll we have a slightly outdated stack. You could also enable Github pages of your fork and check the result there, but as you said there's probably no need to check simple changes.

Ah, that sounds like a good solution. I'll give it a try. Or I might try one of the other methods to get Jekyll running on my machine....

> Oh, did you set up OpenDylan's website? The docs are pleasant to read and the theme is stellar (Sphinx, I see).

I did. I'm quite happy with Sphinx.

One of the ideas kicking around in my head is to try and make it easy to provide better docs for CL projects. Sphinx, with a custom CL domain, would be one way to do that. The docs could have cross references to other CL projects' docs, a multi-project search index, etc.  Sphinx would work, but wouldn't it be cool if we could do it all in Common Lisp?

vindarel:
> with a custom CL domain

did you see https://github.com/russell/sphinxcontrib-cldomain/ ? It works©, I set it up for an unpublished project.

(it wouldn't work for the Cookbook, for which it would be cool to do it all in CL indeed)

> a multi-project search index, etc

that's interesting (reminds me of Ultralisp's symbol search).

cgay:
> > with a custom CL domain
> 
> did you see https://github.com/russell/sphinxcontrib-cldomain/ ? It works©, I set it up for an unpublished project.

Thanks, I'll make a note of it.

[...]

PR #545, opened 2024-06-13
Minor fixes for iteration, style guide, front page, etc

cgay:
applies to #540


vindarel:
that's grand, thanks.

We have to talk about code snippets starting with `* `, I really don't like them: beginners wonder what this is, more tedious to copy-paste, ugly, less explicit than `CL-USER> `. I would not have any prompt (so the styleguide isn't clear on this, but it doesn't mention prompt strings^^).

oh, I see you also uncommented the results, which makes sense when using `* `.

so instead of

```lisp
(map 'vector (lambda (it) (+ it 10)) '(1 2 3))
;; #(11 12 13)
```

we now have

```lisp
* (map 'vector (lambda (it) (+ it 10)) '(1 2 3))
#(11 12 13)
```

this is undoing edits I made for this.

(for long results we can use a `=>` without `;;` afterwards)

cgay:
ok, that makes sense. Just to make sure I understand correctly, you're saying you would prefer

```lisp
(some lisp form)
;; result value 1
;; result value 2
```
?

Or we could go with this, perhaps a little clearer?
```lisp
(some lisp form)
;; => result value 1
;; => result value 2
```

vindarel:
option 1) is used throughout the Cookbook, option 2) is good, we can go with it.

With long results, we can do

```lisp
(some lisp form)
;; =>
;; 1
;; 2
;; 3
```

but also

```lisp
(some lisp form)
;; =>
(some
   result
   form/s)
```

(this allows to see syntax highlighting for the result when appropriate)

and also a separate code block for the result.

cgay:
oh, thanks for merging that. Sorry to have let it slide for so long.

PR #546, opened 2024-06-21
numbers: include the Number Tower

fosskers:
As discussed on Discord. This adds an image to the top of the page that allows the reader to see the relationships between all the main number types at a glance.


numbers.md
```diff
@@ -11,6 +11,14 @@ Some sources:
 * [`Numbers, Characters and Strings`][numbers-characters-strings]
   in Practical Common Lisp
 
+<!-- epub-exclude-start -->
```
fosskers, line 14: I borrowed this pragma from the index page.

numbers.md
```diff
@@ -11,6 +11,14 @@ Some sources:
...
+<div style="text-align: center">
+    <img src="numbertower.png" alt="Number Types in Common Lisp"/>
+</div>
+<!-- epub-exclude-end -->
+
+*Types in bold, cyan boxes are the ones you will typically use.*
```
fosskers, line 20: Is this enough elaboration?

fosskers:
Hi, any thoughts?

kilianmh:
LGTM. It's a good starting point. Is ECL the only implementation with differing specific number type, or are there also others?

fosskers:
Originally the image was much more complicated because it included all the weird hidden subtypes of various things (mainly children of `complex`) that you aren't actually expected to use yourself. I took them back out to make it clearer.

With regards to `long-float`, the precision of floats is implementation dependent, according to the spec: https://cl-community-spec.github.io/pages/short_002dfloat.html

As far as I found, `double-float` and `long-float` are usually the same. ECL in particular makes it extra long, based on C's `long double`.

numbers.md
fosskers: I added more about ECL in the Floating Point section.

fosskers:
Thanks.

kilianmh:
We thank you for your efforts!

PR #547, opened 2024-06-29
iterating on files, lazily

vindarel:
from https://stackoverflow.com/questions/78550959/how-can-i-lazily-traverse-a-directory-in-lisp

### DIFF
--- a/files.md
+++ b/files.md
@@ -693,6 +693,25 @@ Returns a list of pathnames:
  #P"/home/vince/projects/cl-cookbook/assets/")
 ```
 
+#### Iterating on files (lazily)
+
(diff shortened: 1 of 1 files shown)


files.md
```diff
@@ -693,6 +693,25 @@ Returns a list of pathnames:
  #P"/home/vince/projects/cl-cookbook/assets/")
 ```
 
+#### Iterating on files (lazily)
+
+In addition of the above functions, we mention solutions that *lazily*
```
cgay, line 698: In addition *to* ...

vindarel: thanks! done

PR #565, opened 2025-01-09
Minor condition system improvements

phoe:
Clarified and added a few things to the error handling page, cleaned up naming wrt "catching"/"throwing"/"raising".


vindarel:
many thanks

PR #568, opened 2025-02-13
Update pattern_matching.md for association and property list patterns

wmannis:
Added examples for property (and property!), assoc, plist, and alist, which are underdescribed both here and in the trivia docs themselves.


vindarel:
super useful thank you.

PR #571, opened 2025-03-07
amplify section on Closures a bit

wmannis:
Add a little more info about use cases for closures.

Also one trivial formatting fix for pattern matching.

### DIFF
--- a/functions.md
+++ b/functions.md
@@ -652,7 +652,9 @@ Read the CLHS section about [form evaluation](http://www.lispworks.com/documenta
 
 ## Closures
 
-Closures allow to capture lexical bindings:
+Closures allow you to capture lexical bindings. This can be useful to store state
+that you don't want to have to keep passing into your function(s), either as a convenience
+or to keep state variables out of a reachable namespace.
 
 ~~~lisp
 (let ((limit 3)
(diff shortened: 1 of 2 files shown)


vindarel:
Thank you! We need more fresh eyes like yours.

We don't *need* the memo to be a closure, right? It could be a `defvar`. So I'm afraid that the sentence

>  another common use of lexical closures is memoization

is misleading.

```lisp
(defvar *memo* (make-hash-table))

(defun fibonacci (n)
    (let ((value (gethash n *MEMO*)))
      (cond ((<= n 1) n)
            (value value)
            (t (setf (gethash n *MEMO*)
                     (+ (fibonacci (- n 1)) (fibonacci (- n 2))))))))
```

Thoughts?

wmannis:
Before I wrote this I went and checked a few memoization libraries that show up in Quicklisp. Some do use a lexical closure, some use other mechanisms. So, while closures aren't required for this use case, I'm not sure it's misleading to say it is a common use case.

vindarel:
> a hash table in a lexical closure

this association annoys me.

Could you say somehow that the important bit is the HT, using a closure being a common use case / we can use a defvar too?

like

> Using a hash table to store previously caclulated results can speed things up. We can use a `defvar`, but that's a good use case for a closure [and possible re-state why]:

thanks :pray:

wmannis:
Done!

PR #573, opened 2025-03-16
Update data-structures.md - symbol-macrolet with structs

wmannis:
Under structs, mention the use of `symbol-macrolet` as a way to improve readability which is directly analogous to CLOS's `with-slots`.

### DIFF
--- a/data-structures.md
+++ b/data-structures.md
@@ -1965,12 +1965,38 @@ symbol macros which expand into forms with structure accessors:
 ~~~
 
 Here the math involved in the `move-ship` function is easier to
-read than if accessor functions were used.
+read than if accessor functions were used.  Without `symbol-macrolet`
+it looks like this:
+
+~~~lisp
+(defun move-ship (ship)
+  (psetf (ship-x-position ship)
+           (+ (ship-x-position ship) (ship-x-velocity ship))
+         (ship-y-position ship)
+           (+ (ship-y-position ship) (ship-y-velocity ship))))
+~~~
+
+In this function all the accessors are not too hard to read, but
+with more complex operations it would quickly get cluttered.
 
 Though it is not mentioned in the standard, many modern implementations
 of Common Lisp permit the use of the CLOS macro `with-slots` with
 structures.  In the standard `with-slots` itself is defined using
-`symbol-macrolet`.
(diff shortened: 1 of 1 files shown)


data-structures.md
```diff
@@ -1945,6 +1945,33 @@ Slots are `setf`-able:
...
+       (yv (ship-y-velocity ship)))
+    (psetf x (+ x xv)
+           y (+ y yv))))
+~~~
+
+Here the math involved in the `move-ship` function is easier to
+read than if accessor functions were used.
+
+Though it is not mentioned in the standard, many modern implementations
+of Common Lisp permit the use of the CLOS macro `with-slots` with
```
vindarel, line 1971: Can you precisely say which implementations?

vindarel:
Very useful. Would you like to expand a bit more on what it does? (or in another PR if you prefer) Showing the expanded form would be useful, so we don't guess, we see and we ensure we understood. This explanation could even go on a new page that we would add in the index.

data-structures.md
wmannis: I wasn't sure how much to include about this, for three reasons. First, I wasn't sure how much time I wanted to spend installing all the CL implementations that'll run on my mac. Second, the `with-slots` thing fails (on SBCL, at least, and I imagine others) if you change the underlying representation of the struct to a list or a vector (a capability not discussed in this section). Finally, it's not technically part of the standard, even if people do use it.

I could rely on the kindness of strangers on discord to do a few tests for implementations I don't have straightforward access to.

wmannis:
> Would you like to expand a bit more on what it does? (or in another PR if you prefer) Showing the expanded form would be useful, so we don't guess, we see and we ensure we understood.

I wasn't sure how much detail to give, but I agree it makes sense to give something to compare to.

data-structures.md
vindarel: this is also good information that could appear in a subsection : )

\+1 to ask for feedback. Or just mention at least one implementation, for a start.

wmannis:
I'm not sure what good, modern use cases there are for using the `list` or `vector` type representation of structures. I could come up with examples of various option combinations, but I'm not sure it's not historical reenactment.

vindarel:
Thank you!

PR #576, opened 2025-03-19
Update functions.md - multiple values

wmannis:
a nearly complete rewrite of the section on multiple values, trying to highlight use cases based on the CL standard functions that use them

### DIFF
--- a/functions.md
+++ b/functions.md
@@ -283,6 +283,9 @@ Common Lisp has also the concept of multiple return values.
 
 ### Multiple return values
 
+Returning multiple values is *not* like returning a tuple or a list of
+results.
+
 While most Common Lisp forms return a single value, it is
 sometimes useful for a function to return several (or none).
 For example, `round` returns two values, the rounded result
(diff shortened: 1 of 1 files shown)


vindarel:
those are better explanations

I liked my first sentence ("Returning multiple values is not like returning a tuple or a list of results") and the first dummy example with a b c, could your content come after them?

(Don't hesitate to add subsections)

wmannis:
After the fact this occurred to me as a good example, too —

```lisp
(+ (values 1 2 3) (values 10 20 30))
;; => 11
```

wmannis:
I wasn't able to come up with a great way to recover your `a b c` example without including all of it up to `multiple-value-bind`, which doesn't blend with the rest.

wmannis:
In the `(nth-value 0 '(:a :b :c))` examples, the explanatory comments go well beyond the column limit.

vindarel:
> In the `(nth-value 0 '(:a :b :c))` examples, the explanatory comments go well beyond the column limit.

thanks, fixed.

PR #577, opened 2025-03-25
two minor strings updates

wmannis:
1. There were some places where the old uppercase way of referring to FUNCTIONs. was still being used. I converted them to `function`s.
2. In the `format` section I added a quick bit about at-sign and colon modifiers, which are used in the examples, but not specifically called out anywhere.

### DIFF
--- a/strings.md
+++ b/strings.md
@@ -916,7 +916,25 @@ For our examples below, we'll work with a list of movies:
 
 Format directives start with `~`. A final character like `A` or `a`
 (they are case insensitive) defines the directive. In between, it can
-accept coma-separated options and parameters.
+accept coma-separated options and parameters.  Further, some directives
+can take colon and at-sign modifiers, which change the behavior of the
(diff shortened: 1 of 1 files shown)


strings.md
```diff
@@ -916,7 +916,25 @@ For our examples below, we'll work with a list of movies:
...
+;; => "2025"
+(format nil "~:d" 2025)
+;; => "2,025"
+(format nil "~@d" 2025)
+;; => "+2025"
+(format nil "~@:d" 2025)
+;; => "+2,025"
+~~~
+
+If it there isn't a sensible interpretation for both modifiers used
```
vindarel, line 936: if it there isn't ?

vindarel:
Thank you!!

strings.md
wmannis: Ugh. Fixed. I added another directive example, with `~r` keeping with the 2025 theme, but which I forgot.

Issue #590, opened 2025-07-10
Granular build options


nepomucen-sexp:
Hi,

It would be great to have a better control over which files will be build.  Right now `make epub` will build both epub and pdf, which is a bit misleading. Something like this would be awesome:
- `make all` - make all available book formats (epub and pdf for now)
- `make epub` - make epub only
- `make pdf` - make pdf only (I guess by removing the epub file after the pdf is created)

I can create a pull request for that.

What do you think?

vindarel:
Hello, at its core the Cookbook is a static site, so `make all` might be misleading, it doesn't show it's to build documents. The PDF is built from the epub, it seems useless to delete the epub. Can we have `make epub+pdf` of sort? +1 for `make epub` to make epub only.

Issue #589, opened 2025-07-10
Build option to choose the sed program


nepomucen-sexp:
Hi,

I had trouble building epub, because on macOS the default `sed` program was not working correctly. The fix was to install `GNU sed` from `homebrew` and manually change the program name in `make-cookbook.lisp` to `gsed`.  It would be great if the build would support build option to set the `sed` program name through the env variables.

I can create a pull request for that.

Thanks for creating this amazing book!

dotemacs:
I guess this issue can be closed?

PR #592, opened 2025-07-22
Makefile: sed environment variable and granular build options

nepomucen-sexp:
Solves #590 and #589.


vindarel:
Clean and documented, thank you!

PR #593, opened 2025-08-12
strings: add type hierarchy diagram

fosskers:
As I had previously done for Numbers, I've added a type hierarchy for Strings and Chars. I've also spruced up the original Numbers image, and added the `.dot` I used for both diagrams.


vindarel:
very well done

fosskers:
I'll add more as a follow-up. I've already made a collections chart.

PR #594, opened 2025-08-16
data: collections type hierarchy

fosskers:


vindarel:
:fire: A sentence to help the reader appreciate it would still be nice. Like, first the legend.

Issue #587, opened 2025-06-26
Remove Portacle? It is unmaintained and old


vindarel:
warning by its author https://github.com/portacle/portacle/issues/182

> The last time portacle got a proper look was 2020. Since then it has been unmaintained.

> I don't have the patience and energy for this project as Linux and especially Apple continue their war against their own users by breaking everything that used to work fine. And it sure seems like nobody else has the energy for this either, as no one has stepped up to pick the mantle and fix everything that broke in the interim.

Could we bring some $$$ for a newer look?

Would bundling stuff into an Emacs AppImage be that useful? (for linux users) https://github.com/probonopd/Emacs.AppImage/

vindarel:
so, Portacle is an outdated Emacs but it still just works©. What do you all think, should we remove it or add warnings?

PR #597, opened 2025-08-24
Removed portacle & links to it

dotemacs:
As per the issue:
https://github.com/LispCookbook/cl-cookbook/issues/587, portacle is no longer maintained, so removing the links to it.

Also took the opportunity and updated the section on how to install Emacs, git, SBCL & Quicklisp on both macOS & Windows.

Not sure where `windows.md` file is linked from. Figured you might be able to point me, if needed.


vindarel:
IIRC this windows page isn't linked anymore (but left for old links), we put installation instructions in "getting started", since we can refactor some instructions. A sub-section for MacOS and more for Windows would be welcome.

dotemacs:
> thanks. Why, in your opinion, should we remove Portacle? I'd appreciate some discussion. Does it harm users, if we keep it as a solution, but we warn its Emacs version is outdated, and won't give features x y and z ?

My logic was:
- project is stale
- the author won't maintain it
- why give users a stale project to "help" them

I suspect that being given the stale project, won't give much confidence in the book itself.

But again, this is just my view. Happy to go with whatever you decide.

dotemacs:
I've revered the update to the `windows.md` file, as per comment above, since it's no longer used.

I've added instructions on how to install Emacs.

I've updated the instructions on how to install `rlwrap` for all three major OSs.

See what you make of it.

Thanks

Issue #306, opened 2020-02-26
Website overhaul?


vindarel:
Continuing the nascent discussion in #279:

we'd enjoy a faster website generator (for local work), and a better looking theme.

Hugo themes for documentation: https://themes.gohugo.io/tags/documentation/

Use Coleslaw? https://github.com/coleslaw-org/coleslaw

Mkdocs theme: https://squidfunk.github.io/mkdocs-material/ (python)

vindarel:
Searching alternatives to gitbook: https://alternativeto.net/software/gitbook/?license=opensource

- https://rust-lang.github.io/mdBook/ same look, experimental EPUB generator
- https://daux.io/Getting_Started.html multi-language support. Can build a single HTML file, so probably OK for PDF generation. Example in the wild: https://vulkan-tutorial.com/
- …

pnathan:
I like this idea.

I also like the idea of coleslaw, because it's in lisp. But that's got no technical bearing, I think.

arademaker:
Well, racket website is made in racket. We can’t ignore the impact of this.

vindarel:
> Well, racket website is made in racket. We can’t ignore the impact of this.

at least common-lisp.net, Quickdocs and Cliki are made with CL… 

---

I stumbled on [Docsify](https://docsify.js.org/#/quickstart) and tried it: https://vindarel.github.io/cl-cookbook/

I like it and it seems to cover all our needs. It's a full-JS rendering engine that renders the markdown files, we don't have to generate HTML.

What's still there:

- the TOC on the left sidebar, same structure overall

What's new:

- modern, fashion-compliant look and feel
- a welcome screen (it is currently the default)
- a search box (although I experienced a couple issues)

What's missing and could be added:

- syntax highlighting with HighlightLisp, which is better
- a link to the page's source (either with an existing plugin either with our proper code)
- we might loose on Search Engine Optimization, but it's possible to use a Server Side Rendering plugin.

What's possible:

- a dark theme with a switch button
- link to translations page per page
- multi-language support
- more plugins: https://docsify.js.org/#/plugins (edit on github, copy to clipboard button, offline reading,…), https://docsify.js.org/#/awesome

What's impossible:

- to read an url like `…/#/getting-started` with Emacs' Web Browser… but it's possible to read `/getting-started.md` in markdown! We could maybe advertise the alternative url on every page.

---

What do you all think, any strong objections?

I'll keep this open a few days.

[...]

PR #600, opened 2025-09-11
Refactor HTML/SCSS and improve performance, accessibility, and TOC handling

hwahyeon:
This PR got a bit bigger than expected 😅

- Replace deprecated `escape()` with `encodeURIComponent()`
> Reference: [MDN Web Docs - escape()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape)

- Remove deprecated `type="text/javascript"` from script tags
> Since Netscape 2, the default scripting language in browsers has been JavaScript, so `type="text/javascript"` is no longer required.

- Add `preconnect/dns-prefetch` hints for faster external resource loading
> Reference: [MDN Web Docs - rel="preconnect"](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/preconnect)
> Reference: [MDN Web Docs - rel="dns-prefetch"](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/dns-prefetch)
- Change search input type from `text` to `search`
- Replace hacky form action with proper event handling in the search form
- Add hidden `<label>` to the search input for web accessibility
- Add `defer` attribute to head scripts to avoid render blocking
- Fix `duckSearch` function to prevent page reload
- Replace hardcoded `/cl-cookbook/` path with a variable
- Format HTML for readability
- Fix TOC not visible on the index page
image

- Move TOC/column inline script to an external file with `defer`
- Remove duplicate comments in SCSS

### DIFF
--- a/assets/style.scss
+++ b/assets/style.scss
@@ -181,7 +181,6 @@ footer {
 }
 
 .announce-green {
-  // background-color: MediumAquamarine
   background-color: MediumAquamarine;
 }
 


vindarel:
Many thanks!

If you are motivated, we could use a newer static site with a better theme! See #306 (only need, we need not break the page URLs, or at least have a redirection).

vindarel:
on init-toc.js, there is still

```ls
// Don't write the TOC on the index.
if (window.location.pathname !== "{{ site.baseurl }}/") {
```

which isn't required anymore (since you put the TOC on the index).

hwahyeon:
@vindarel  My mistake! I overlooked that part. I've fixed it in PR #601, so the TOC is now hidden again on the index page. If you think it would be better to have the TOC visible on the index page, please let me know.

PR #601, opened 2025-09-14
Fix TOC showing on index page and HighlightLisp error

hwahyeon:
- Hide TOC on the index page
(regression from previous PR #600 ) 
- Add guard so TOC init runs only if jQuery, the plugin, and `#toc` exist
(not strictly necessary, but for safety)  
- Run `highlight_auto()` after `highlight-lisp.js` to fix ReferenceError


vindarel:
Please add comments on init-toc.js to say

    // Don't write the TOC on the index.

and to say ` && $ && $.fn && $.fn.toc` is for safety and what it checks 

and we're golden!

(suppose a non-web dev reads the js : ) )

PR #602, opened 2025-09-24
Edited for conciseness

### LINKED ISSUE #506: Sockets tutorial goes on tangents, used `unwind-protect` instead of macros
The sockets tutorial is kind of a mess to read. Both sections are weighed down with an extended tangent about the "mistakes" that the author encountered, which are confusing to read and can't be much more helpful than simply explaining how to use the library correctly.

Additionally, the code on display ignores the convenience macros that usocket offers, such as  instead implementing equivalent functionality with `unwind-protect`, requiring that the reader think about conditions and control flow when they should be focusing on the socket API.

This section should be rewritten to be more clear and direct, and should prefer the `with-` macros defined by the usocket package to `unwind-protect`.

I'll probably give a shot at rewriting it soon.

NicoSimoski:
Sorry, it took me some time because of several distractions on my way...

- removed statements about mistakes when prototyping
- used with- macros supplied by usocket

Closes [#506](https://github.com/LispCookbook/cl-cookbook/issues/506)

### DIFF
--- a/sockets.md
+++ b/sockets.md
@@ -6,14 +6,14 @@ This is a short guide to TCP/IP and UDP/IP client/server programming in Common
 Lisp using [usockets](https://github.com/usocket/usocket).
 
 
-## TCP/IP
+## How do I create a TCP/IP server and let a client send data to it?
 
 As usual, we will use quicklisp to load usocket.
 
     (ql:quickload "usocket")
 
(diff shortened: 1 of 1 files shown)


vindarel:
excellent, many thanks for the contribution to the Cookbook, that's really helpful.

PR #603, opened 2025-09-30
Revised and extended the section on plists

dotemacs:
I've changed the symbols to keywords, for keys, in the existing example.

Then extended the section on how to append to plists and explained the difference in the approaches. And I've also added a section on naming of the keys.


dotemacs:
Added the subsections, as per your suggestion @vindarel 

Let me know what you think, thanks

PR #618, opened 2026-03-14
Add streams chapter

### LINKED ISSUE #542: Feature Request: Chapter on streams
Hi.

I think a chapter on streams would be a good addition. The "Input/Output" section contains some stuff related to streams, but there's more to cover.

I'm creating this issue for future reference. I'm considering trying to create the chapter myself.

The Common Lisp the Language book has a good chapter on streams, there's also the Gray Streams, etc.

kovan:
New chapter covering the Common Lisp stream system:

- **Stream basics** — types, element types, predicates
- **Standard streams** — `*standard-input*`, `*standard-output*`, etc. with table
- **File streams** — `open`, `with-open-file`, binary files
- **String streams** — `with-output-to-string`, `with-input-from-string`, `make-string-output-stream`
- **Concatenated streams** — reading from multiple streams in sequence
- **Broadcast streams** — writing to multiple streams simultaneously
- **Two-way and echo streams** — bidirectional and logging streams
- **Synonym streams** — stream indirection via symbols
- **Gray streams** — extending the stream protocol with CLOS, using `trivial-gray-streams`

All examples are self-contained and runnable. Added to the index under "Language basics" after Input/Output.

Fixes #542

### DIFF
--- a/streams.md
+++ b/streams.md
@@ -1,0 +1,304 @@
+---
+title: Streams
+---
+
+Streams are the standard abstraction for input and output in
+Common Lisp. Every time you read from a file, write to the
+terminal, or communicate over a network socket, you are using
+a stream. This chapter covers the stream types, how to create
+and use them, and how to extend the stream protocol.
+
+For basic read/write recipes, see also
+[Input/Output](io.html).
+
+## Stream basics
+
+A stream is an object that represents a source or sink of
+characters or bytes. The standard defines several stream
+types:
+
+- **Input streams** support reading (`read-char`,
+  `read-byte`, `read-line`, `read`).
+- **Output streams** support writing (`write-char`,
+  `write-byte`, `write-string`, `format`).
+- **Bidirectional streams** support both.
+- **Character streams** carry characters (the default).
+- **Binary streams** carry unsigned bytes, specified by
+  an element type like `(unsigned-byte 8)`.
+
+You can test what a stream supports:
+
+~~~lisp
+(input-stream-p *standard-input*)   ;; => T
+(output-stream-p *standard-output*) ;; => T
+(stream-element-type *standard-input*)
+;; => CHARACTER
+~~~
+
+## Standard streams
+
+Common Lisp provides several global stream variables that
+are bound by default:
+
+| Variable | Purpose |
+|---|---|
+| `*standard-input*` | Default input (your terminal) |
+| `*standard-output*` | Default output (your terminal) |
+| `*error-output*` | Error/warning messages |
+| `*trace-output*` | Output from `trace` |
+| `*debug-io*` | Interactive debugging I/O |
+| `*query-io*` | User yes/no questions |
+| `*terminal-io*` | The actual terminal stream |
+
+Functions like `read`, `print`, and `format` use these by
+default when you don't specify a stream:
+
+~~~lisp
+;; these are equivalent:
+(print "hello")
+(print "hello" *standard-output*)
+~~~
+
+You can rebind them with `let` to redirect output:
+
+~~~lisp
+(let ((*standard-output* some-other-stream))
+  (print "hello"))
+  ;; prints to some-other-stream
+~~~
+
+## File streams
+
+Use `open` to create a file stream, or the
+`with-open-file` macro which ensures the stream is
+properly closed:
+
+~~~lisp
+;; reading a file:
+(with-open-file (stream "/tmp/test.txt")
+  (loop for line = (read-line stream nil)
+        while line
+        do (print line)))
+~~~
+
+~~~lisp
+;; writing to a file:
+(with-open-file (stream "/tmp/out.txt"
+                 :direction :output
+                 :if-exists :supersede)
+  (format stream "Hello, streams!~%"))
+~~~
+
+The `:direction` keyword controls the stream type:
+
+- `:input` (default) — read only
+- `:output` — write only
+- `:io` — read and write
+- `:probe` — just check if the file exists, then close
+
+For binary files, specify `:element-type`:
+
+~~~lisp
+(with-open-file (stream "/tmp/data.bin"
+                 :direction :output
+                 :if-exists :supersede
+                 :element-type '(unsigned-byte 8))
+  (write-byte 72 stream)
+  (write-byte 101 stream))
+~~~
+
+## String streams
+
+String streams let you treat strings as streams, which is
+useful for building output or parsing input without files.
+
+### Writing to a string: `with-output-to-string`
+
+~~~lisp
+(with-output-to-string (s)
+  (format s "Hello, ~a!" "world"))
+;; => "Hello, world!"
+~~~
+
+This is the idiomatic way to build strings with
+`format`, `write-string`, or other stream operations.
+
+### Reading from a string: `with-input-from-string`
+
+~~~lisp
+(with-input-from-string (s "123 456")
+  (list (read s) (read s)))
+;; => (123 456)
+~~~
+
+### `make-string-input-stream` and `make-string-output-stream`
+
+For cases where the macro forms are inconvenient, you
+can create string streams directly:
+
+~~~lisp
+(let ((s (make-string-output-stream)))
+  (format s "one ")
+  (format s "two ")
+  (format s "three")
+  (get-output-stream-string s))
+;; => "one two three"
+~~~
+
+~~~lisp
+(let ((s (make-string-input-stream "hello")))
+  (read-char s))
+;; => #\h
+~~~
+
+## Concatenated streams
+
+`make-concatenated-stream` creates a stream that reads
+from multiple input streams in sequence. When the first
+stream is exhausted, reading continues from the next:
+
+~~~lisp
+(let* ((s1 (make-string-input-stream "Hello, "))
+       (s2 (make-string-input-stream "world!"))
+       (combined (make-concatenated-stream s1 s2)))
+  (read-line combined))
+;; => "Hello, world!"
+~~~
+
+## Broadcast streams
+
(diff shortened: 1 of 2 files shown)


vindarel:
This chapter is needed.

I can't review properly until ±7 days, however I think some things would need to be more explained.

Please, what is your usage of AI for this PR? Does this contribution come 100% from CL experience?

kovan:
Thanks for the initial look. Happy to expand the explanations — if you can point out which parts feel underexplained when you do the full review, I'll address them.

Re AI: I used Claude as a drafting aid and then edited and verified against the HyperSpec. The structure and topic selection are mine. Happy to discuss if you have concerns about the workflow.

streams.md
```diff
@@ -0,0 +1,304 @@
...
+String streams let you treat strings as streams, which is
+useful for building output or parsing input without files.
+
+### Writing to a string: `with-output-to-string`
+
+~~~lisp
+(with-output-to-string (s)
+  (format s "Hello, ~a!" "world"))
+;; => "Hello, world!"
+~~~
```
vindarel, line 126: Tell this is equivalent to `format nil`. 

Also we could give as example a `print-object` method that uses a `stream` argument.

streams.md
```diff
@@ -0,0 +1,304 @@
...
+This is the idiomatic way to build strings with
+`format`, `write-string`, or other stream operations.
+
+### Reading from a string: `with-input-from-string`
+
+~~~lisp
+(with-input-from-string (s "123 456")
+  (list (read s) (read s)))
+;; => (123 456)
+~~~
```
vindarel, line 162: An explanation and an idea of use case would be nice.

streams.md
```diff
@@ -0,0 +1,304 @@
...
+types:
+
+- **Input streams** support reading (`read-char`,
+  `read-byte`, `read-line`, `read`).
+- **Output streams** support writing (`write-char`,
+  `write-byte`, `write-string`, `format`).
+- **Bidirectional streams** support both.
+- **Character streams** carry characters (the default).
+- **Binary streams** carry unsigned bytes, specified by
+  an element type like `(unsigned-byte 8)`.
```
vindarel: I think we can split this group:

- input/output/bidirectionnal
and
what streams read by default:
- character
- bytes

streams.md
```diff
@@ -0,0 +1,304 @@
...
+~~~
+
+This is the idiomatic way to build strings with
+`format`, `write-string`, or other stream operations.
+
+### Reading from a string: `with-input-from-string`
+
+~~~lisp
+(with-input-from-string (s "123 456")
+  (list (read s) (read s)))
```
vindarel, line 160: Shall we recall / explain `read` here? (and `read-char` below?)

streams.md
```diff
@@ -0,0 +1,304 @@
...
+
+The standard stream types are implemented by the
+Common Lisp runtime. If you need custom stream behavior
+(for example, a stream that compresses data, counts
+bytes, or transforms characters), you can use
+**Gray streams**.
+
+Gray streams are a de facto standard (supported by SBCL,
+CCL, ECL, ABCL, LispWorks, Allegro, and others) that
+lets you define stream classes with CLOS methods.
```
vindarel: More history?

The Gray proposal was made before the ANCI CL standard was finalized,
and was based on the Common Lisp The Language book.

The proposal did not make it into ANSI CL, but most popular
CL implementations implement this facility anyway.

streams.md
```diff
@@ -0,0 +1,304 @@
...
+       (counting (make-instance 'counting-stream
+                                :inner out)))
+  (write-string "hello" counting)
+  (values (get-output-stream-string out)
+          (char-count counting)))
+;; => "hello"
+;; => 5
+~~~
+
+The key methods to implement depend on the stream type:
```
vindarel, line 393: Their README first tells about STREAM-READ-SEQUENCE and STREAM-WRITE-SEQUENCE, let's also mention them.

A bit more about what they do would be welcome.

streams.md
```diff
@@ -0,0 +1,304 @@
...
+This lets you redirect where a stream goes by rebinding
+the symbol, without changing the stream object itself.
+
+## Gray streams: extending the protocol
+
+The standard stream types are implemented by the
+Common Lisp runtime. If you need custom stream behavior
+(for example, a stream that compresses data, counts
+bytes, or transforms characters), you can use
+**Gray streams**.
```
vindarel, line 348: More of the why, "why can we not use the standard stream types for this" seems important.

notes for myself:

> The purpose of gray streams is to integrate with Common Lisp, and to allow you to pass them around between applications transparently. Thus the support is geared towards the CL standard read-char/byte/sequence. 

(shinmera on reddit)

streams.md
```diff
@@ -0,0 +1,304 @@
...
+
+- `stream-write-char` — write one character
+- `stream-line-column` — current column (or `nil`)
+- `stream-write-string` (optional, for performance)
+
+**Binary streams:**
+
+- `stream-read-byte`
+- `stream-write-byte`
+
```
vindarel, line 419: Question: what's the equivalent of sending to `/dev/null` already? A good example to add.

streams.md
```diff
@@ -0,0 +1,304 @@
...
+You can test what a stream supports:
+
+~~~lisp
+(input-stream-p *standard-input*)   ;; => T
+(output-stream-p *standard-output*) ;; => T
+(stream-element-type *standard-input*)
+;; => CHARACTER
+~~~
+
+## Standard streams
```
vindarel: "variables" should be in the title I guess

https://cl-community-spec.github.io/pages/Stream-Variables.html

streams.md
```diff
@@ -0,0 +1,304 @@
...
+**Binary streams:**
+
+- `stream-read-byte`
+- `stream-write-byte`
+
+## Further reading
+
+- [CLHS: Streams](http://www.lispworks.com/documentation/HyperSpec/Body/21_.htm)
+- [CLtL2: Streams](https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node329.html)
+- [trivial-gray-streams](https://github.com/trivial-gray-streams/trivial-gray-streams)
```
vindarel, line 424: Mention https://edicl.github.io/flexi-streams/ ?

also https://github.com/yitzchak/nontrivial-gray-streams ?

what about https://franz.com/support/documentation/10.1/doc/streams.htm "Besides Gray streams, Allegro CL has its own set of stream classes, simple-streams, another interesting design: " (ruricolist)

streams.md
```diff
@@ -0,0 +1,304 @@
...
+;; => 5
+~~~
+
+The key methods to implement depend on the stream type:
+
+**Character input streams:**
+
+- `stream-read-char` — read one character
+- `stream-unread-char` — push a character back
+- `stream-read-line` (optional, for performance)
```
vindarel, line 400: also STREAM-READ-CHAR-NO-HANG ?

streams.md
kovan: Addressed in `0bace12`. I added a short history note here covering the pre-ANSI origin and the fact that Gray streams remained a de facto standard.

streams.md
kovan: Addressed in `0bace12`. I split stream direction from element type so input/output/bidirectional and character/binary are presented separately.

streams.md
kovan: Addressed in `0bace12`. I expanded this section with more explanation and a concrete use case for creating string streams directly when the macro forms are inconvenient.

streams.md
kovan: Addressed in `0bace12`. I added the `format nil` comparison here and included a `print-object` example that writes to the supplied stream.

streams.md
kovan: Addressed in `0bace12`. I added a short explanation of `read` here and contrasted it with `read-char` below.

streams.md
kovan: Addressed in `0bace12`. I added further-reading links for `flexi-streams`, `nontrivial-gray-streams`, and Allegro simple-streams.

streams.md
kovan: Addressed in `0bace12`. I expanded the motivation here to explain why Gray streams are needed when you want custom stream classes to participate in ordinary CL I/O.

streams.md
kovan: Addressed in `0bace12`. I added the `make-broadcast-stream` no-argument example as the portable `/dev/null`-style sink.

streams.md
kovan: Addressed in `0bace12`. I added `stream-read-char-no-hang` to the character-input method list.

streams.md
kovan: Addressed in `0bace12`. I added `stream-read-sequence` and `stream-write-sequence` here and noted why they matter for bulk reads and writes.

streams.md
kovan: Addressed in `0bace12`. I renamed the section to `Standard stream variables`.

vindarel:
thanks. It would be nice if we exposed a real-world example. Do you guys have an idea?

streams.md
```diff
@@ -226,6 +226,67 @@ that stream is discarded.
...
+;; => ==============
+;; => Books: 12.50 EUR
+;; => Train: 24.10 EUR
+;; => Lunch: 18.00 EUR
+;; => --------------
+;; => Total: 54.60 EUR
+;; => "
+~~~
+
+If you want tee-style output, you can also combine destinations
```
vindarel: "tee-style": let's recall what it is ("tee: Copy standard input to each FILE, and also to standard output", where the important bit is writing to 2 output streams)

thanks

vindarel:
not bad, not bad. One last remark.

PR #622, opened 2026-03-14
Add map-into examples to iteration chapter

### LINKED ISSUE #442: Feature Request: Explanation with examples of map-into
http://www.lispworks.com/documentation/HyperSpec/Body/f_map_in.htm

kovan:
Adds `map-into` examples after the `map`/`mapcar` section, showing usage with lists and vectors. Includes a note that it's destructive and that `map`/`mapcar` should be preferred when allocation is not a concern, per @vindarel's guidance.

Fixes #442

### DIFF
--- a/iteration.md
+++ b/iteration.md
@@ -143,6 +143,30 @@ expects the type for its result as first argument:
 ;; => "abc"
 ~~~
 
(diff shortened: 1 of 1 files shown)


vindarel:
Thanks!

PR #635, opened 2026-07-26
chat/ not needed in URI

JohnConner:
Just a simplification to avoid any confusion - websocket doesn't need a URI to a webpage.


vindarel:
thank you

PR #636, opened 2026-08-31
Add an action to auto build the pdf/epub

pvonmoradi:
This workflow builds and releases the pdf/epub outputs on each tagged commit.
Originally, I had it generate a draft release on each commit push and overwrite previous outputs since it would provide a pdf/epub that is in sync with the web version (like [this](https://github.com/pvonmoradi/cl-cookbook/releases)). 
This PR only does a release on each tag.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,53 @@
+name: Build and release the book pdf and epub
+on:
+  push:
+    tags: ['*']
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Prepare the dev environment
+        run: |
+          sudo apt update
+          sudo apt install build-essential sbcl cl-quicklisp calibre
+          # install lisp library deps
+          yes '' | sbcl --non-interactive \
+            --eval '(load "/usr/share/common-lisp/source/quicklisp/quicklisp.lisp")' \
+            --eval '(quicklisp-quickstart:install)' \
+            --eval '(ql:add-to-init-file)' \
+            --eval '(ql:quickload "str")'
+          # install a newer pandoc
+          wget https://github.com/jgm/pandoc/releases/download/3.10.2/pandoc-3.10.2-1-amd64.deb
+          sudo apt install ./pandoc-3.10.2-1-amd64.deb
+
+      - name: Download and install Typst
+        run: |
+          wget -O- "https://github.com/typst/typst/releases/latest/download/typst-x86_64-unknown-linux-musl.tar.xz" \
+            | tar -xJ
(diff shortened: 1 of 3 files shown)


vindarel:
Thank you!

- we'll want the year and date in the released file: `common-lisp-cookbook-2026-01.pdf`
- we have a date to change in metadata.txt (January 2026 is hardcoded)
- we have to update contributors.md which includes the content of commiters.md which, is the comment is correct, is generated with `git shortlog -sn | cut -f 2 | sed s/^/"* "/g > commiters.md`. There's room for improvement…

Can you do that?

(so I guess these automation improvements would go in the Makefile)

pvonmoradi:
> we'll want the year and date in the released file: `common-lisp-cookbook-2026-01.pdf`

Handled this in the action file.

> we have a date to change in metadata.txt (January 2026 is hardcoded)
> 
> we have to update contributors.md which includes the content of commiters.md which, is the comment is correct, is generated with `git shortlog -sn | cut -f 2 | sed s/^/"* "/g > commiters.md`. There's room for improvement…
> (so I guess these automation improvements would go in the Makefile)

I think since these changes need to be committed to the repo, we shouldn't use an automated script to non-transparently mutate the files on release without commits. So I thought a `make` target that just checks whether these files are updated for release is the better approach.
I added a check-release function in the lisp build script. It uses regexp to extract year/month and contributors. But it seem complicated or rather brittle to me...

vindarel:
Okay.

Thank you!

vindarel:
I tagged today's release as pre-realease. https://github.com/LispCookbook/cl-cookbook/releases/tag/2026-09-12

Looks like page 3 "2. Content" (the index.md ?) shouldn't be here.

pvonmoradi:
Thanks for following up. Two issues:

1- The `make check-release` is not run anywhere in the action file. That's why the build runs without triggering errors when `metadata.txt` and `contributors.md` are not updated.
We can add `make check-release` before `make epub+pdf` so the build fails and the maintainer can be alerted that he needs to manually commit changes to those files.

2- Can you review the `check-release` function in `make-cookbook.lisp`? I'm new to CL and I think the implementation is shaky and regexp-dependent. I don't know how these "pre-commit" checks are usually handled in projects like this. The `git shortlog ...` hint comment in `contributors.md` shows piping to a `commiters.md` but this file does not exist in repo.

vindarel:
I'm fine with not running `check-release` in the CI. A thing less to debug in "prod". It will serve as a developer tool.

You could maybe use local-time (new home: https://github.com/sharplispers/local-time) but that's ok.

The main issue for automatic publication is to not include the index.md in the PDF build. I surely did this manually last time, since it was the first release with Typst.

PR #637, opened 2026-08-31
fix call-arguments-limit

pvonmoradi:
On SBCL, the call-arguments-limit is either (ash 1 30) on x86_64 or array-dimension-limit on non-x86_64 architectures. https://github.com/sbcl/sbcl/blob/master/src/compiler/early-c.lisp#L18-L29


pvonmoradi:
Apparently on ARM64 of Apple Silicon, array-dimension-limit is 4611686018427387903 .

